PR #1, opened 2021-11-23
Improvments to the Launcher.

edfloreshz:
I replaced GtkEntry with GtkSearchEntry to allow the search field to be cleared with a button.
image

I think I figured out a way to clear GtkListBox without a loop, let me know your thoughts @jackpot51.

I also added a new commit to include the launcher example in Cargo.toml and update the README.md.


edfloreshz:
Added a new commit to include the launcher example in Cargo.toml and update the README.md

PR #2, opened 2021-11-24
feat: Use IpcClient from pop-launcher-service

mmstick:
- Improve launcher example to asynchronously handle to pop-launcher responses using the IpcClient API in pop-launcher
- pop-launcher's own event loop will be spawned locally on the glib main context.
- pop-launcher responses and GTK signals will send their events to an event loop on the glib main context.

Requires https://github.com/pop-os/launcher/pull/65 to be merged in advance


jackpot51:
Obsoleted by #6

PR #4, opened 2021-11-29
WIP: Gtk4 Application Launcher Prototype

wash2:
Application launcher prototype using GTK4 templates and GIO AppInfo.

The gtklauncher example currently supports searching applications, launching applications, shortcuts, and closing the window when it is deselected.

The existing IPC based launcher prototype is also changed to a partially implement a ListStore of custom GObjects that wrap the IPC SearchResult.


jackpot51:
Obsoleted by #6

PR #15, opened 2022-09-30
Made the navigation bar scrollable 

edfloreshz:
- Added padding to the entries to avoid collisioning with the scrollbar


jackpot51:
This appears to break the responsive layout when the window has a small width

edfloreshz:
True, let me fix that.

edfloreshz:
Let me know your thoughts 👍🏼

mmstick:
All seems to be working well with these changes

PR #17, opened 2022-10-08
Widget implementation

edfloreshz:
This PR includes implementation of various widgets, including: 

- [x] NavBar
- [x] Expander


mmstick:
Looks good to me

mmstick (approved):

mmstick:
I've been working on a similar approach for the headerbar widgets.

edfloreshz:
I noticed this morning, I'm testing to see if I can implement something similar for this widget, I'll let you know.

edfloreshz:
@mmstick  I'm facing an issue when trying to implement that approach

In order to follow the builder pattern and be able to render the elements inside the navigation bar I need to store them in the struct:

```rust
pub fn with_children(children: Vec<Element<'a, Message, Renderer>>) -> Self {
    Self {
        children,
        ..Default::default()
    }
}
```

But when trying to create a `Column` from that vector in the `render` method I'm getting the following error:

```bash
    |
129 | Column::with_children(self.children)
    |                       ^^^^^^^^^^^^^ move occurs because `self.children` has type `Vec<iced_native::Element<'_, Message, Renderer>>`, which does not implement the `Copy` trait
```

Can you think of a way to overcome this?

mmstick:
You won't be able to store elements in the application's struct. They will have to be recreated for each call of the view method.

edfloreshz:
I figured I could add the children parameter to the render method and it works.

I'm currently working on an Expander widget and it even looks nice.

mmstick:
I'm not fully sold on the API of the headerbar in the current source code. Exploring something closer to what Iced is doing, but keeping the Headerbar data struct since it nicely abstracts state unique to the headerbar widget.

mmstick:
Sounds like we'll have a full widget library in no time at all

edfloreshz:
> I'm not fully sold on the API of the headerbar in the current source code. Exploring something closer to what Iced is doing, but keeping the Headerbar data struct since it nicely abstracts state unique to the headerbar widget.

Alright, let me know when you have something. 👍🏼

edfloreshz:
By the way, we can merge this if Jeremy approves, once it's merged I'll make another PR to adopt the new API style.

edfloreshz:
I made some changes to the `HeaderBar` widget.

I added a generic type `Message` to the struct and added closures to send data to the parent. 

```rust
pub struct HeaderBar<Message>
{
    title: String,
    nav_title: String,
    sidebar_active: bool,
    show_minimize: bool,
    show_maximize: bool,
    on_sidebar_active: Box<dyn Fn(bool) -> Message>,
    on_close: Box<dyn Fn() -> Message>,
    on_drag: Box<dyn Fn() -> Message>,
}
```


 instead of using the `Widget` trait, I utilized the `Component` trait, this allowed me to handle events for the component and send events to the parent when necessary.

```rust
match event {
    HeaderEvent::Close => Some((self.on_close)()),
    HeaderEvent::ToggleSidebar => {
        self.sidebar_active = !self.sidebar_active;
        Some((self.on_sidebar_active)(self.sidebar_active))
    },
    HeaderEvent::Drag => Some((self.on_drag)()),
    HeaderEvent::Minimize => None,
    HeaderEvent::Maximize => None,
}
```

It is no longer necessary to store the widget inside the parent struct, we just need to create it and pass the closures:

```rust
header_bar(
    self.title().as_str(),
    self.sidebar_toggled,
    self.show_minimize,
    self.show_maximize,
    |toggled| AppMsg::ToggleSidebar(toggled),
    || AppMsg::Close, 
    || AppMsg::Drag 
);
```

Then we can handle it in the parent like so:

```rust
fn update(&mut self, message: Self::Message) -> Command<Self::Message> {
    match message {
        AppMsg::Close => self.exit = true,
        AppMsg::ToggleSidebar(toggled) => self.sidebar_toggled = toggled,
        AppMsg::Drag => return drag(),
        AppMsg::Minimize => {},
        AppMsg::Maximize => {},
        AppMsg::Page(page) => self.page = page,
    }
    Command::none()
}
```

I also added a new example called `settings` in which I plan to start building the widgets we discussed earlier.

I believe this could be the way to build future widgets, let me know your thoughts @mmstick

mmstick:
Yeah, this is exactly what I had in mind for the headerbar.

mmstick:
Although I think it's still better to have a `HeaderData` struct store in the application's struct so that the function using it doesn't have to take so many parameters, which may change over time as more features are added.

mmstick:
Going to merge so I can iterate on this

edfloreshz:
I'm implementing that next.

Instead of having all the events in the enum, I was thinking of doing something like this: 

```rust
enum AppMsg {
    Header(HeaderMsg)
}
```

To distinguish the events of other components and avoid rewriting them in the parents enum, like you were doing before.

PR #20, opened 2022-10-14
Updated Navigation Bar

edfloreshz:


jackpot51:
Could you describe what these updates do?

examples/text/src/font/fallback/mod.rs
```diff
@@ -36,7 +36,12 @@ pub struct FontFallbackIter<'a> {
 }
 
 impl<'a> FontFallbackIter<'a> {
-    pub fn new(fonts: &'a [Font<'a>], default_family_opt: Option<&'a str>, scripts: Vec<Script>, locale: &'a str) -> Self {
+    pub fn new(
```
jackpot51, line 39: I do not want the format of this code changed

edfloreshz: Sorry, I ran `rustfmt` on the whole project. I'll change it back.

PR #23, opened 2022-10-15
View Switcher Widget

edfloreshz:
This PR implements the View Switcher widget.

image
image


edfloreshz:
Was there an issue with #20, #21 or #23?

PR #25, opened 2022-10-29
Include Cosmic theme in libcosmic, instead of in Iced fork

ids1024:
I think it's best to only include things in the Iced fork that can't be done without forking Iced, and/or are expected to be merged upstream. The theme doesn't seem to be either, so it fits more appropriately in libcosmic.

That should make it easier to keep up with upstream, and it may help to have all the Cosmic parts in one place.

Based on https://github.com/pop-os/iced commit ad9026e.


mmstick (approved):

jackpot51 (approved):
This is a good idea.

PR #39, opened 2022-12-16
refactor: allow the icon to be non-svg

### LINKED ISSUE #38: Need icon loader that works with svg and image
Icons don't load properly in https://github.com/pop-os/cosmic-applets/pull/33 since applications tend not to offer their icons as svgs. While `Button::icon()` and thus `CosmicAppletHelper::icon_button()` only loads svgs.

I think for this sort of use, we'd want something that loads an icon as either a bitmap or svg, favoring symbolic icons, and if applicable themes the icon? We may want to do this for all icons rather than requiring svgs anywhere, since users may want to use icon themes that don't provide icons as svgs. Even if alternate icon themes aren't "supported" they should mostly work.

There should also be some fallback if an icon fails to load.

wash2:
This allows an icon to be non-svg. It would fix #38

### DIFF
--- a/src/widget/icon.rs
+++ b/src/widget/icon.rs
@@ -7,13 +7,13 @@ use iced::{
     widget::{svg, Image},
     Length, ContentFit,
 };
-use std::borrow::Cow;
+use std::{borrow::Cow, ffi::OsStr};
 use std::hash::Hash;
 use std::rc::Rc;
 use derive_setters::Setters;
 use crate::{Element, Renderer};
 
-/// A lazily-generated SVG icon.
+/// A lazily-generated icon.
 #[derive(Hash, Setters)]
 pub struct Icon<'a> {
     #[setters(skip)]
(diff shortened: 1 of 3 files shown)


src/widget/icon.rs
```diff
@@ -57,41 +46,59 @@ pub fn icon<'a>(name: impl Into<Cow<'a, str>>, size: u16) -> Icon<'a> {
...
             };
 
-            let mut widget = svg::Svg::<Renderer>::new(handle)
-                .style(svg.style)
-                .width(svg.width.unwrap_or(Length::Units(svg.size)))
-                .height(svg.height.unwrap_or(Length::Units(svg.size)));
-
-            if let Some(content_fit) = svg.content_fit {
-                widget = widget.content_fit(content_fit);
+            if icon.as_ref().map(|path| path.extension() == Some(&OsStr::new("svg"))).unwrap_or(true) {
```
mmstick, line 68: Perhaps we could also use https://crates.io/crates/infer to get the file type in the event that the file has no extension

src/widget/icon.rs
```diff
@@ -30,18 +30,7 @@ pub struct Icon<'a> {
     height: Option<Length>,
 }
```
mmstick, line 31: Maybe we could add a `force_svg: bool` property so the caller could optionally avoid the check.

src/widget/icon.rs
```diff
@@ -30,18 +30,7 @@ pub struct Icon<'a> {
...
-        .map(|path| {
-            Image::new(path)
-                .width(Length::Units(size))
-                .height(Length::Units(size))
-        })
-}
-/// A lazily-generated SVG icon.
+/// A lazily-generated icon.
 #[must_use]
 pub fn icon<'a>(name: impl Into<Cow<'a, str>>, size: u16) -> Icon<'a> {
```
mmstick, line 35: Another idea: we'll definitely want to be able to load an icon from an embedded image, so perhaps the `name` property should be something more precise. An `IconSource` type.

src/widget/icon.rs
wash2: I think that would be great, but as far as I can tell, there is no method of checking specifically for svg in the infer crate :/

mmstick (approved):

PR #40, opened 2022-12-17
feat: aspect ratio limiting container

wash2:
This adds a container that limits its contents to the specified aspect ratio.


mmstick (approved):

mmstick:
This one might be a good idea to RFC to Iced.

PR #49, opened 2022-12-22
More efficently resize at break point

13r0ck:


jackpot51 (commented):
It would be best to only send a message when the window condensed state changes, otherwise, I think this recalculates the view on every change in size as before.

13r0ck:
Because Iced subscriptions require a `fn` not a closure, we can't pass the app state into `events_with`. What we could do is modify the [window resize event](https://github.com/iced-rs/iced/blob/d2d18479acd987b4a4f6a125c0dda7b2a6661df4/winit/src/application.rs#L209) to include the previous window size, then the `fn` could check if the change between new and old crosses the break point

13r0ck:
Well that's just cool. It works how you want it too now, and I'm smarter too!

examples/cosmic/src/window.rs
```diff
@@ -258,6 +266,22 @@ impl Application for Window {
...
+          if old_width == 0
+            || old_width < BREAK_POINT && width > BREAK_POINT
+              || old_width > BREAK_POINT && width < BREAK_POINT {
+            Some(width)
+          } else {
+            None
+          }
+        }
+        _ => None
+      }).map(Message::WindowWidthChanged)
```
jackpot51, line 282: Since this message is only set when going past the break point, I'd recommend just sending a message that only has a bool for condensed or not, instead of sending the width.

13r0ck: Good point. I did one small change, I set the static atomic u32 in the subscription, so that I don't actually have to send any data, and just sent a unit type.

jackpot51 (approved):

examples/cosmic/src/window.rs
jackpot51: Perfect!

PR #50, opened 2022-12-23
Fix builds for cosmic-applets

wash2:
This fixes the builds for cosmic-applets according to my testing https://github.com/pop-os/cosmic-applets/commit/da00e9f7c12cd97b8b45908e9f100a4fe95d597b


mmstick (approved):

ids1024:
Ah, did this end up pulling in two different versions of `cosmic-panel-config` somewhere? `cargo vendor` doesn't seem to like that for git dependencies.

PR #62, opened 2023-01-03
refactor!: separate horizontal and vertical segmented button widgets

mmstick:
- Removes the orientation enum in favor of two separate widgets
- Implements the spacing attribute for both widgets
- Demo is updated to display spaced variants of the widgets

Screenshot from 2023-01-03 19-55-45
Screenshot from 2023-01-03 19-55-57


13r0ck (approved):
Works for me! Thank you!

PR #64, opened 2023-01-03
Basic CI

Drakulix:
- First commit adds a github actions file for testing the different features of libcosmic as well as the examples and some rustfmt checks
- Second commit formats the whole repo. If this is controversial we can easily drop the check and this commit, but I would personally welcome enforcing this.
- Third commit updates iced to https://github.com/pop-os/iced/pull/18 (pending) and fixes up the features to always enable a suitable shell+renderer combination

(Actually enabling usage without any shell for cosmic-comp will be done in a later PR.)


mmstick (dismissed):

mmstick:
All checks passing. If that's all we need for now, then we can merge

wash2 (approved):

PR #66, opened 2023-01-05
Global setting for default icon theme

### LINKED ISSUE #61: Don't force the icon-theme to be "Pop"
We are currently forcing the icon-lookup to search for the pop-os theme: https://github.com/pop-os/libcosmic/blob/master/src/widget/icon.rs#L80

The `freedesktop-icons` crate will exclusively search for icons in that theme, falling back to "hicolor".
If the pop theme is not installed, this can cause most icons to be missing including quite important ones such as "window-minimize-symbolic" or "window-maximize-symbolic", which are not included in "hicolor" (at least not on Fedora).

While we are building up our own design language, I think we should play nice here and pick up the icon theme from either the set `GTK_THEME` (if any) or adjust `freedesktop-icons` to allow more fallback options.

(Additionally we want some placeholder icon for missing ones, I guess, but that should probably be a separate issue.)

mmstick:
Applications may now override the default icon theme application-wide using a new convenience function. Typically called before running the application.

```rs
cosmic::settings::set_default_icon_theme("Pop");
```

It can be applied at any moment during the lifetime of the application. The cosmic example has been updated to demonstrate the ability to dynamically switch between loading icons from the Pop and Adwaita themes. Ideally we'll want to support loading this setting from a configuration file of some kind, too.

Closes #61

### DIFF
--- a/examples/cosmic-sctk/src/main.rs
+++ b/examples/cosmic-sctk/src/main.rs
@@ -7,6 +7,7 @@ mod window;
 pub use window::Window;
 
 pub fn main() -> cosmic::iced::Result {
+    settings::set_default_icon_theme("Pop");
     let mut settings = settings();
     settings.initial_surface = InitialSurface::XdgWindow(Default::default());
     Window::run(settings)
--- a/examples/cosmic/src/main.rs
+++ b/examples/cosmic/src/main.rs
@@ -7,9 +7,8 @@ mod window;
 pub use window::*;
 
 pub fn main() -> cosmic::iced::Result {
+    settings::set_default_icon_theme("Pop");
     let mut settings = settings();
     settings.window.min_size = Some((600, 300));
-    // TODO: Window resize handles not functioning yet
-    settings.window.decorations = false;
     Window::run(settings)
 }
--- a/examples/cosmic/src/window.rs
+++ b/examples/cosmic/src/window.rs
@@ -19,7 +19,7 @@ mod bluetooth;
 
 mod demo;
 
-use self::{demo::DemoView, desktop::DesktopPage};
+use self::desktop::DesktopPage;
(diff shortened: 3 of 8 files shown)


Drakulix:
This still means, that "Pop" would be the unconditional default in all of our apps. I question that default. Don't we want to have some sensible fallback logic, if icons are not available? Or even integrate into the users desktop by respecting their theme settings?

mmstick:
The only sensible defaults at the moment are Adwaita and Pop. Anything beyond that will need some form of configuration file support.

Drakulix:
> Hicolor is missing all the icons we're using in the settings app.

Right, which is why I would suggest cascading defaults. So modifying the `icon`-widget in a way, that we can check another theme, if the first one doesn't result in anything.
That way we could for example query first "Pop" and then "Adwaita".

> KDE, Mate, GNOME, etc. all set icon themes differently, too. GNOME uses gsettings get org.gnome.desktop.interface icon-theme

Long term we need a settings for this or some abstraction to query this on different DEs.

mmstick:
It'd be best if only freedesktop-icons manages fallbacks, since any fallbacks we define may have already been searched. The Pop icon theme defines that it `Inherit`s Adwaita and Hicolor, so it's already recursively searching multiple themes there.

Drakulix:
> It'd be best if only freedesktop-icons manages fallbacks, since any fallbacks we define may have already been searched. The Pop icon theme defines that it `Inherit`s Adwaita and Hicolor, so it's already recursively searching multiple themes there.

In that case this is an issue for freedesktop-icons given that the inheritance is clearly not applied on my system.

Drakulix (approved):
Code-wise this looks fine to me! :+1:

PR #70, opened 2023-01-06
Navigation bar reimplemented with the vertical view switcher

mmstick:
- Adds some ergonomic methods for constructing its states
   - `State::insert_active` activates the inserted button
   - `State::builder()` makes a builder that can add multiple buttons
- Adds icon support to the segmented button widget
- Replaced the nav bar macros used by the cosmic example with the view switcher widget
    - Renamed instances of `sidebar` with `nav_bar`

Screenshot from 2023-01-06 02-04-57


examples/cosmic/src/window.rs
```diff
@@ -121,23 +125,24 @@ const BREAK_POINT: u32 = 900;
...
-    theme: Theme,
     bluetooth: bluetooth::State,
+    debug: bool,
     demo: demo::State,
     desktop: desktop::State,
-    system_and_accounts: system_and_accounts::State,
-    sidebar_toggled: bool,
-    sidebar_toggled_condensed: bool,
-    show_minimize: bool,
+    nav_bar_pages: segmented_button::State<Page>,
```
13r0ck, line 132: Since Iced has gone 'stateless' holding widget state in the application state It probably not what we should be doing

13r0ck: Maybe something like, the widget could accept an array of `("icon name", "button label", message_when_clicked)`?
Then I don't think the state manipulation would be needed

mmstick: That would make the widget more stateful rather than less. The only state that each instance of the widget holds internally is the ID of the button being hovered by the mouse, because a mouse can only hover over one unique instance of a widget at a time. Beyond that, the widget doesn't actually hold any state outside of configuration parameters for drawing.

Here the application maintains all application state rather than the widget. It has sole ownership over its state to update it in the update method, and supplies an immutable reference to the widget in the view. There's not a single shared reference or any interior mutability.

This makes it easy for the application to keep track of which buttons are active or enabled. It doesn't require any comparison checks to determine that in the view or the update method. It also saves us a lot of resources since the strings and paths don't have to be recreated in every view update. Having a unique ID in the message makes it quick to remove or modify state related to the button.

mmstick:
Multi-select now supported, depending on which variant of a state is used. `State<SingleSelect>` has single-select behavior whereas `State<MultiSelect>` has multi-select.

[MultiSelect.webm](https://user-images.githubusercontent.com/4143535/211042201-c7e7c588-91fe-4b88-8aec-4c9f1b4eee4c.webm)

13r0ck (changes requested):
Sorry, I know elm style widgets are very different than gtk style widgets, I get why you want to do it this way.  Manipulating widget state (even if that state is held in the application state ) shouldn't be allowed. Particularly the setting a widget to multi-select or not via manipulating a state. 

In traditional elm style this should just be helper function to just reorganize existing widgets. But a widget is fine too, it just means more work. For example this will not work with iced's tab focus system without the other buttons being implemented as children.

If this were a helper function it would take a `vec` of what should be displayed there. Which would just be buttons, or the content of the buttons. I know you are looking for optimizations, so maybe we could have a `const` array of values that the widget can then enumerate into this widget?

mmstick:
> Sorry, I know elm style widgets are very different than gtk style widgets, I get why you want to do it this way.

This would be very strange to someone familiar with GTK, because it's a data-oriented approach with stateless widgets utilizing message-passing in an event loop with one source of truth in the application's model. This comes more from my experience with ECS in Rust rather than GTK. It's much closer to ECS and Elm than an object-oriented approach.

In GTK, almost all application state is held internally by each widget. A typical Rust application using GTK has very little state in the application, and potentially even no state at all. Rather, the widgets contain most of your state, and you can only influence their state externally. They're shared pointers with interior mutability; with many, many aliases of these widgets passed around as weak references to various callbacks so that they can get the data from a widget or interact with it. 

As far as I can tell, this is exactly what the Elm style recommends. To keep the model separate from the view. To pass the model by immutable reference in the view method to generate the view. For the view to emit messages back to the event loop, and for those messages to passed into the update method where the model can mutate its state based on the message received.

That's precisely what's happening here. This is no different from passing a `&str` into a widget's function. I've just encapsulated the necessary data for the model into a single type so that we don't require someone using the widget to have to reimplement all of the logic by hand, and that lets us achieve some relatively complex behaviors with a small amount of code. Whether that's tracking active and enabled status, components associated with those IDs, or even reordering them.

Similar to what happens when you pass a `&str` or `String`, this model is passed into the widget's function in the view, and it's only holding onto an immutable reference to the model that you pass into it for a brief moment. Same as when you are storing that string in the widget's struct. At no point in runtime does one of these widgets actually have the power to manipulate state. Only the application can do this in the `update` method, typically when it receives a message that a button was activated.

In the application's model, a `SlotMap` is used to generate a unique integer ID on insert, where part of that ID is a generation ID, so that slots can be safely reused, unlike a Slab which cannot. When this is given to the widget, it's simply iterating on the slot map to get the contents to generate the widget, and the ID to emit on button activation.

Unlike with GTK, there aren't any shared pointers or interior mutability here. The widget never manipulates the input it receives. It's just getting the IDs and contents of each button, and using message-passing to convey the ID of the activated button back to the application.

Then it's up to the application to decide how to respond to that message. Maybe you disable the button so that it cannot be clicked again. Run an external command in the background to do the thing it activates. Then when you get the response from that command, you can decide to enable the button and activate or deactivate it.

> Particularly the setting a widget to multi-select or not via manipulating a state.

The state is not being manipulated (unless you mean the compile-time generation of the state type?). Technically, it is two separate model types, and four separate segmented button types that are generated. The alternative approach would be using enums to switch between modes, which is what I think you're thinking this doing. Similar to how we dealt with having two separate types for orientation rather than using an enum, I've chosen the same approach here so a person can opt between the two different selection models based on the type declared or the function used that declares the type.

You could create type aliases and functions for them as I already did with `HorizontalSegmentedButton = SegmentedButton<Horizontal>` and `VerticalSegmentedButton = SegmentedButton<Vertical>`. With the `horizontal_segmented_button()` returning a `SegmentedButton<Horizontal>`, and the cosmic function `horizontal_view_switcher()` returns a pre-configured `SegmentedButton<Horizontal>` with the `ViewSwitcher` style.

With MultiSelect as an option, these are now possible types:

- `SingleSelectHorizontalSegmentedButton`
- `SingleSelectVerticalSegmentedButton`
- `MultiSelectHorizontalSegmentedButton`
- `MultiSelectVerticalSegmentedButton`

But that's a bit wordy, and Rust can thankfully express that with a state machines:

- `SegmentedButton<Horizontal, SingleSelect>`
- `SegmentedButton<Vertical, SingleSelect>`
- `SegmentedButton<Horizontal, MultiSelect>`
- `SegmentedButton<Vertical, MultiSelect>`

Which made this especially easy to implement since the widget only needs

Which type is generated is based on which model type you pass into the `vertical_` or `horizontal_` function of the widget. So passing a `State<SingleSelect>` into a `horizontal_view_switcher()` creates a `SegmentedButton<Horizontal, SingleSelect>`.

Any behaviors common between them can them be implemented generically. Whereas any behavior specific to a variant can be implemented separately, such as how there are separate implementations for the horizontal and vertical segemented buttons, and separate implementations for the single and multi-select models.

```rs
impl State<SingleSelect> {}
impl State<MultiSelect> {}
impl SegmentedButton<Horizontal, Selection> {}
impl SegmentedButton<Vertical, Selection> {}
```

> For example this will not work with iced's tab focus system without the other buttons being implemented as children.

I can work on implementing that. It is possible already to manage focus for each button separately by tracking which button has focus by its ID. Similar to how it tracks which button is actively being hovered by storing it in the `Widget::State`. I can expose them as children if necessary.

> If this were a helper function it would take a vec of what should be displayed there. Which would just be buttons, or the content of the buttons. I know you are looking for optimizations, so maybe we could have a const array of values that the widget can then enumerate into this widget?

This would be identical to what's already being done once you implement all of the features required. It's a SlotMap rather than a Vec, and the caller only needs to define a single message variant to interact with it, but the idea is the same. I think you're mainly having issue with constructing the initial model in the Default impl, rather than constructing the model in the view. You mentioned a const array of values, for example, but that will only work for a static interface with a finite number of values that are already known at compile-time. I'm thinking more about dynamic interfaces with tabs in a text editor or web browser, or the ability to populate a list from data from an external source. I can support a const array of values for initializing the model, but I opted for the functional approach with a builder. I've just not figured out a better way to declare the contents of a button in this way, though I've toyed with an insert method that returns a builder rather than inserting immediately.

PR #75, opened 2023-01-16
Warning widget

edfloreshz:
This PR implements a warning widget, it's just a banner that shows text and a close button, it uses the COSMIC theme warning colors.

If there's any change in styling of functionality, please let me know. 👍🏼 

image
image


src/widget/warning.rs
```diff
@@ -0,0 +1,69 @@
...
+use std::borrow::Cow;
+
+use iced::{alignment, widget, Alignment, Background, Color, Length};
+
+use crate::{theme, Renderer, Theme};
+
+use super::icon;
+
+#[must_use]
+pub fn warning<Message>(message: &str) -> Warning<Message> {
```
Drakulix, line 13: You could make the message argument a `impl Into<Cow<'a, str>>` and it would accept both, borrowed `str`s and owned `String`s.

edfloreshz: Sure

edfloreshz: Should be resolved now 👍🏼

mmstick:
@pop-os/ux Do we have a design for a warning/infobar widget at the moment? Would these be the correct colors?

maria-komarova:
Haven't finished the design for that one yet. And technically - yes, would be correct colors. We ran into one complication so can end up changing them because of it. But the complication is only in regards to the button on top of the notification so the colors should work for now.

mmstick (approved):

PR #76, opened 2023-01-18
feat!(segmented-button): improved interfaces and documentation

mmstick:
Changes made while working on improving the ergonomics of the code in the Settings prototype.

BREAKING CHANGE: Various type and function names have changed to reflect
themselves better in documentation. Code has been reorganized into
separate modules with a better placement in libcosmic. Most of the
functions, types, and modules now have documentation and examples.

These changes no longer require the `Model` type to define the
data/component type that it stores. The component functionality is now
optional, and it's also possible to associate many components to an item
with one component per type. This has had a side effect of simplifying a
lot of the type signatures in the implementation.

Before, to insert an item into the model, you had to define a
`SegmentedItem` and a `Component` on insert, and get back an ID for that
item. Which makes it difficult to define an item that contains only an
icon or has no components. And requires an extra insert function to
activate the item on insert.

```rs
let a_id = model.insert_activate("Item A", 0);
let b_id = model.insert(segmented_button::item().icon("custom-icon"), 1);
```

Now, there is a flexible builder-style API for configuring
newly-inserted items in the model. The `SegmentedItem` no longer exists,
preferring to store the text and icons in separate stores. So the complexity
for inserting and retrieving values from the model has decreased
significantly.

```rs
let a_id = model.insert().text("Item A").data(0).activate().id();
let b_id = model.insert().icon("custom-icon").id();

if let Some(text) = model.text(id) {
    println!("this button has {text}");
}

if let Some(icon) = model.icon(id) {
    println!("this button has an icon: {:?}", icon);
}
```

Equivalent changes to the API for the `segmented_button::Model::builder()`
have also been applied, as seen in the demo example.

---

A side effect of the model type no longer requiring to define a `Component`
is that it's now possible to store any number of secondary data components
to an item, provided that each component has a unique Type ID. The built-in
data store is a TypeId-based HashMap with a secondary map for each registered type.

```rs
struct App {
    pub nav_items: segmented_button::SingleSelectModel
}

let mut items = segmented_button::SingleSelectModel::default();

let id = nav_items.insert()
    .text(page.title())
    // associate a u16 data component
    .data(0u16)
    // associate a Page data component
    .data(page)
    .id();

// Fetches the static string data component
if let Some(page) = items.data::<Page>(id) {
    println!("found page: {}", page.title());
}

// Fetches the u16 data component
if let Some(number) = items.data::<u16>() {
    println!("found number: {}", number);
}
```

Finally, an example for associating data externally from the model.

```rs
let mut secondary_data = segmented_button::SecondaryMap::default();


let page_id = model.insert().text(page.title());
secondary_data.insert(page_id, page);

// Or it may be inserted functionally with
model.insert()
    .text(page.title())
    .secondary(&mut secondary_data, page);
```

---

This also adds a new method for setting the position of an item by an index.

```rs
model.position_set(id, 0);
```

Which is also supported with the new insertion API.

```rs
let c_id = model.insert().text("Item C").position(0).id();
```

---

I've also experimented with supporting an alternative model input that you can construct in the view within a `Vec`, but while it worked it was not very intuitive to manage, nor as efficient, so I scrapped it for now. It looked like this:

```rs
let choose_icon_theme = horizontal_segmented_selection(vec![
    segmented_button::vec_item()
        .text(&self.icon_themes[0])
        .active(self.icon_theme_active == 0),
    segmented_button::vec_item()
        .text(&self.icon_themes[1])
        .active(self.icon_theme_active == 1),
])
.on_activate(Message::IconTheme);

match message {
    Message::IconTheme(key) => {
        self.icon_theme_active = key;
        if let Some(theme) = self.icon_themes.get(self.icon_theme_active) {
            cosmic::settings::set_default_icon_theme(*theme);
        }
    }
}
```


jackpot51:
@13r0ck please review

13r0ck (dismissed):
This is an improvement on what was implemented in #70

PR #82, opened 2023-01-30
cosmic-settings improvements

mmstick:
Changes applied while developing cosmic-settings

- Adds `widget::text()` function that takes `Cow<str>` as input for improved performance over `iced::widget::text()` that takes `impl ToString`. As `Text::new()` actually takes a `Cow<str>` as input.
- Fix for keyboard focus subscription conflicting with text input focus
- Adds Escape and Search messages for keyboard navigation subscription
- Refactor to add divider style variant functions
- Adds a `Search` style for the TextInput widget
- Fixes handling of empty titles in headerbar
- Fixes spacing of packed widgets in headerbar
- Adds the search widget used by cosmic-setings
- Refactors SingleThreadExecutor to the `executor` module and sets an alias of `executor::Default` to it. Also adds a `multi::Executor` to explicitly opt into that when the tokio feature is enabled.
- Makes font size configurable in segmented button
- Fixes some of the font sizes that were not defined
- Adds an item builder for the settings section widget


jackpot51 (commented):
The changes look good to me after skimming but I'd like @13r0ck to look at them

wash2 (approved):

src/widget/search/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+//!
+//! ```ignore
+//! let search_field = search::search(&self.search, Message::Search);
+//! ```
+//!
+//! Handle messages in the update method:
+//!
+//! ```ignore
+//! match message {
+//!     Message::Search(search::Message::Activate) => {
```
13r0ck, line 26: I have no issues with this, but this specific `Activate` message seems unnecessary, or at least should be eventually. Focus should generally be handled automatically, with overrides being set by widget operation. 
But the global widget focus isn't built yet. I believe @wash2 is working on that for a11y

More of a heads up that this implementation will have to change in the future, rather than me having any issues with if :slightly_smiling_face:

13r0ck (approved):
> Adds widget::text() function that takes Cow<str> as input for improved performance over iced::widget::text() that takes impl ToString. As Text::new() actually takes a Cow<str> as input.

There was discussion about this [on the iced discord](https://discord.com/channels/628993209984614400/1066035737818046614), and Upstream disagreed with this. Though I agree with mmstick here.


> Fix for keyboard focus subscription conflicting with text input focus

See other comment. Just a heads up that the focus implementation in iced will be changing, so this will have to be re-written again. But no worries for now!

13r0ck:
Sorry about the wait

src/widget/search/mod.rs
wash2: > I have no issues with this, but this specific `Activate` message seems unnecessary, or at least should be eventually. Focus should generally be handled automatically, with overrides being set by widget operation. But the global widget focus isn't built yet. I believe @wash2 is working on that for a11y
> 
> More of a heads up that this implementation will have to change in the future, rather than me having any issues with if slightly_smiling_face

I actually haven't been making changes to the existing focus implementation. I've made additions to Operation today so that i can use the exiting find_focused operation internally. As far as I can tell, this was the method most preferred.

13r0ck: Ah ok, sorry I forgot that discussion was via DM

> My basic idea is some sort of centralized focus system. Instead of every focusable widget having an is_focused flag in its state, the idea is to have a single variable focused_widget: Option<widget::Id> somewhere and provide it to the Widget trait.
> The main challenge here is identifying widgets without an explicit Id provided from the user.
So that continuity is maintained.
\- Hecrj

So the plan is to do that *eventually*, whatever you have for a11y works for now

wash2: Ok gotcha :)

PR #83, opened 2023-02-13
Configurable close icons for view switching

mmstick:
These changes extend the functionality of the segmented button widget to support optional and configurable use of close icons which can emit close messages on click. As well as making it possible to override the default color assigned to an item's icon.

An example is also included, and the `IconSource` enum is refactored to unify the `Embedded` and `EmbeddedSvg` variants as a `Handle` variant, using the existing `icon::Handle` enum.

---

Adds support for displaying and handling tab closes, as well as displaying the close icon only on tabs which are hovered or active.

```rs
view_switcher::horizontal(&app.tabs)
    .show_close_icon_on_hover(true)
    .on_activate(Message::TabActivate)
    .on_close(Message::TabClose)
```

Items will display a close icon only if the closable property is set.

```rs
let id = self.tabs.insert().closable().id();
```

By default, `window-close-symbolic` is being used as the close icon, but it can be changed with

```rs
view_switcher::horizontal(&app.tabs)
    .close_icon(IconSource::from("window-close-symbolic"))
```

To activate the item adjacent to the active item on remove, this can be used:

```rs
if self.tabs.is_active(id) {
    if let Some(pos) = self.tabs.position(id) {
        let next = if pos == 0 { pos + 1 } else { pos - 1 };
        self.tabs.activate_position(next);
    }
}

self.tabs.remove(id);
```

The color of an icon can be modified per-item. By default, icons will be filled with the same color as the text. This behavior can be modified by specifying a color (or lack of color):

```rs
let id = self.tabs.insert()
    .closable()
    .icon_color(None)
    .id();
```

Or with

```rs
self.tabs.icon_color_set(id, None);
self.tabs.icon_color_remove(id);
```

When the color is unset, it will match the same color as the text. When set to `None`, it will not modify the color. And setting it to `Some` color will use that color instead of the default.

### DIFF
--- a/examples/cosmic/Cargo.toml
+++ b/examples/cosmic/Cargo.toml
@@ -10,3 +10,4 @@ apply = "0.3.0"
 fraction = "0.13.0"
 libcosmic = { path = "../..", default-features = false, features = ["debug", "winit_softbuffer"] }
 once_cell = "1.15"
+slotmap = "1.0.6"
--- a/examples/cosmic/src/window.rs
+++ b/examples/cosmic/src/window.rs
@@ -28,6 +28,8 @@ mod demo;
 use self::desktop::DesktopPage;
 mod desktop;
 
+mod editor;
+
 use self::input_devices::InputDevicesPage;
 mod input_devices;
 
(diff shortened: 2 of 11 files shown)


wash2 (commented):
This looks good to me, I guess the only nit-pick i have is that there is no indication of whether the close icon is being hovered / pressed or not. I would expect it to act just like a button widget.

mmstick:
We should be able to improve it to use our theme's button styles

wash2 (approved):

jackpot51 (approved):

PR #85, opened 2023-02-18
Add `_with_flags()` variants to settings helpers

ids1024:


src/applet/mod.rs
```diff
@@ -81,20 +81,26 @@ impl CosmicAppletHelper {
 
     #[must_use]
     pub fn window_settings<F: Default>(&self) -> Settings<F> {
-        let mut settings = crate::settings();
+        self.window_settings_with_flags(F::default())
+    }
```
mmstick, line 85: Perhaps this could be a `impl Default for Settings<F> where F: Default`?

ids1024: `Settings` is a type re-exported from Iced. And this particular method provides the settings for an applet window.

PR #86, opened 2023-02-21
(theme): use cosmic-theme for all widgets

wash2:
This modifies libcosmic to use cosmic-theme internally.


13r0ck (commented):
Is there any reason for us to call it `cosmic-container`? I would rather just import a single container widget

wash2:
> Is there any reason for us to call it `cosmic-container`? I would rather just import a single container widget

I called it CosmicContainer because it is deferent and not meant to be used all the time, only really to denote a layer. Normal containers are still useful, but I'm open to another name than CosmicContainer.

13r0ck:
Why is it not meant to be used all the time?

13r0ck:
I think it might be more ergonomic to just overwrite the container import all together and just have our cosmic `container` widget just have a method to denote layer

wash2:
> I think it might be more ergonomic to just overwrite the container import all together and just have our cosmic `container` widget just have a method to denote layer

Ok I could probably do that

13r0ck (approved):
Thank you!!!

jackpot51 (approved):

PR #89, opened 2023-03-05
Config API

jackpot51:
This adds the Config API as discussed on Pop!_OS chat. The API backend details are abstracted and should be reviewed separately from the frontend.

Tested on Linux, macOS, and Windows. Windows required patching out the use of `freedesktop-icons`, we may want to offer an alternative to that.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -76,6 +76,7 @@ optional = true
 
 [workspace]
 members = [
+  "cosmic-config",
   "examples/*",
 ]
 exclude = [
--- a/cosmic-config/Cargo.toml
+++ b/cosmic-config/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "cosmic-config"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+atomicwrites = "0.4.0"
+dirs = "4.0.0"
+notify = "5.1.0"
+ron = "0.8.0"
+serde = "1.0.152"
--- a/cosmic-config/src/lib.rs
+++ b/cosmic-config/src/lib.rs
@@ -1,0 +1,250 @@
+use notify::Watcher;
+use serde::{de::DeserializeOwned, Serialize};
+use std::{
+    fs,
+    io::Write,
+    path::{Path, PathBuf},
+    sync::Mutex,
+};
+
+#[derive(Debug)]
+pub enum Error {
+    AtomicWrites(atomicwrites::Error<std::io::Error>),
+    InvalidName(String),
+    Io(std::io::Error),
+    NoConfigDirectory,
+    Notify(notify::Error),
+    Ron(ron::Error),
+    RonSpanned(ron::error::SpannedError),
+}
+
+impl From<atomicwrites::Error<std::io::Error>> for Error {
+    fn from(f: atomicwrites::Error<std::io::Error>) -> Self {
+        Self::AtomicWrites(f)
+    }
+}
+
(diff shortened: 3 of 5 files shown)


src/config.rs
```diff
@@ -0,0 +1,208 @@
...
+        }
+        Ok(())
+    }
+}
+
+// Setting any setting in this way will do one transaction for all settings
+// when commit finishes that transaction
+impl<'a> ConfigSet for ConfigTransaction<'a> {
+    fn set<T: Serialize>(&self, key: &str, value: T) -> Result<(), Error> {
+        //TODO: sanitize key (no slashes, cannot be . or ..)
```
mmstick, line 198: Perhaps the key could be a pre-sanitized newtype so that subsequent accesses would not have to repeat the process?

jackpot51: The cost of key_path is low, no filesystem access is required to make sure it ends up a direct child of path

src/config.rs
```diff
@@ -0,0 +1,208 @@
...
+#[must_use = "Config transaction must be committed"]
+pub struct ConfigTransaction<'a> {
+    config: &'a Config,
+    //TODO: use map?
+    updates: Mutex<Vec<(PathBuf, String)>>,
+}
+
+impl<'a> ConfigTransaction<'a> {
+    /// Apply all pending changes from ConfigTransaction
+    //TODO: apply all changes at once
```
ids1024, line 180: Is it possible to replace multiple files at once in an atomic manner on typical filesystems?

It could be done with an SQLite based backend.

jackpot51: It is possible, with symlinks and temporary directories.

jackpot51: And even if it can't be done atomically, it can be done transactionally, for example by having a marker file indicate when a transaction is in progress, and the watcher must see it be removed before sending file updates.

jackpot51: I'm still not sold on sqlite as a backend due to the large amount of C code it would bring in, and the difficulty of managing settings in `git`. There is not much it can offer beyond easy transactions, which are not difficult to set up in a filesystem.

ids1024: Apparently `renameat` with `RENAME_EXCHANGE` can swap two directories atomically? Not sure about on other platforms. In contrast to plain `rename` which moves things into a directory rather than atomically replacing. It would still be desirable to have some kind of locking to make sure multiple processes aren't trying to perform transactions on entries in the same directory.

Directory replacement also changes what filesystem notifications it's necessary to monitor for. Which should be doable though anything like this requires testing on any OS being targeted.

> And even if it can't be done atomically, it can be done transactionally, for example by having a marker file indicate when a transaction is in progress, and the watcher must see it be removed before sending file updates.

That's probably good enough, though SQLite's journaling provides additional protection by making sure the a transaction is never half applied even in the case of a crash.

ids1024:
We'll want to think about the best higher-level API for use in typical iced applets. That fits with the model-view-update paradigm.

A typical Iced application likely has a certain number of settings it needs the values of, and these values should logically be part of the model. There should be a `subscription` to detect changes to settings, which sends a message to `update`, which likely just stores updated values in the model. And then the `view` will use the new values when it is next run.

What's the best way to handle this with minimal boilerplate? If there are 5 settings (for instance), it would be a bit verbose to have to add all 5 to the app's model and deal with all 5 in `update`, possibly after listing all 5 in the subscription.

We also need some way to handle this for any settings used internally by libcosmic.

PR #93, opened 2023-04-05
fix: window configured sizes fix

wash2:


wash2:
I noticed this issue while updating applets. They rely on being able to request a size, but after the update to latest, iced-sctk was setting the size to the defaults instead of checking requested size first.

13r0ck (approved):
After many hours of review. I believe this is ok :stuck_out_tongue:

PR #103, opened 2023-05-08
Cosmic advanced text

wash2:
This is a WIP update to use https://github.com/pop-os/iced/pull/41

Font loading and a few other items are missing, but the examples build and run again.


src/lib.rs
```diff
@@ -5,8 +5,6 @@
 
 pub use cosmic_theme;
 pub use iced;
-pub use iced_lazy;
-pub use iced_native;
```
ids1024, line 9: It may be useful to have re-exports of the new crates: `iced_widget` and `iced_runtime`.

mmstick (dismissed):

mmstick (approved):

PR #109, opened 2023-05-31
refactor: move applet helpers to cosmic-applets repo

wash2:


mmstick (dismissed):
Less default dependencies good

PR #125, opened 2023-07-03
config: Ignore some filesystem events

ids1024:
APIs like inotify can provide notification for some events that don't change the contents of a file. It makes sense to ignore those here, even if the expected way to write is through the config API.

This reduces the number of events on the directory received by a single `:w` in vim from 12 to 8. And reduces a `touch` from 2 events to 1.

Atomic writes through the config API only result in 1 event per-setting, before and after this change.

### DIFF
--- a/cosmic-config/src/lib.rs
+++ b/cosmic-config/src/lib.rs
@@ -2,7 +2,10 @@
 use iced_futures::futures::channel::mpsc;
 #[cfg(feature = "subscription")]
 use iced_futures::subscription;
-use notify::{RecommendedWatcher, Watcher};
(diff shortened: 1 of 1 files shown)


mmstick (approved):

mmstick:
Does this fix an event being emitted when writing the same value to the file?

ids1024:
Writing the same value that already exists in the file? I think on the notification end, it would have to store the value of each key to compare in order to do that.

On the writing end (if the setting is written through cosmic-config), we could read the current value and compare before attempting to write, but that isn't free.

PR #139, opened 2023-08-08
wip: add applet module

wash2:
This seems to address the issues from https://github.com/pop-os/cosmic-applets/pull/110 but maybe a separate trait could be added which adds the style by default.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ winit_tokio = ["iced/winit", "iced_winit", "tokio"]
 winit_wgpu = ["winit", "wgpu"]
 # Enables XDG portal integrations
 xdg-portal = ["ashpd"]
+applet = ["wayland", "tokio", "a11y", "cosmic-panel-config", "ron"]
 
 [dependencies]
 apply = "0.3.0"
@@ -97,6 +98,15 @@ optional = true
 [dependencies.iced_wgpu]
 path = "iced/wgpu"
 optional = true
+
+[dependencies.cosmic-panel-config]
+git = "https://github.com/pop-os/cosmic-panel"
+optional = true
+
+[dependencies.ron]
+version = "0.8"
+optional = true
+
 
 [workspace]
 members = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,8 @@ winit_tokio = ["iced/winit", "iced_winit", "tokio"]
 winit_wgpu = ["winit", "wgpu"]
 # Enables XDG portal integrations
 xdg-portal = ["ashpd"]
-applet = ["wayland", "tokio", "a11y", "cosmic-panel-config", "ron"]
+# XXX Use "a11y"; which is causing a panic currently
+applet = ["wayland", "tokio", "cosmic-panel-config", "ron"]
 
 [dependencies]
 apply = "0.3.0"
--- a/src/app/applet/mod.rs
+++ b/src/app/applet/mod.rs
@@ -128,7 +128,7 @@ impl CosmicAppletHelper {
         &self,
         icon_name: &'a str,
     ) -> iced::widget::Button<'a, Message, Renderer> {
-        crate::widget::button(theme::Button::Secondary)
+        crate::widget::button(theme::Button::Text)
             .icon(theme::Svg::Symbolic, icon_name, self.suggested_size().0)
             .padding(8)
     }
--- a/src/app/applet/mod.rs
+++ b/src/app/applet/mod.rs
@@ -1,0 +1,292 @@
+use std::sync::Arc;
+
+use crate::{
+    app::Core,
+    cosmic_config::CosmicConfigEntry,
+    cosmic_theme::util::CssColor,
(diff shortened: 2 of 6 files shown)


src/app/applet/mod.rs
```diff
@@ -0,0 +1,292 @@
...
+            parent_size: None,
+            grab: true,
+        }
+    }
+
+    #[must_use]
+    pub fn theme(&self) -> theme::Theme {
+        match self.background {
+            CosmicPanelBackground::ThemeDefault | CosmicPanelBackground::Color(_) => {
+                let Ok(helper) = cosmic_config::Config::new(
```
ids1024, line 205: Don't we already have a `system_theme` that should contain this?

wash2: ya i'll update this

cosmic-theme/Cargo.toml
```diff
@@ -12,21 +12,16 @@ rustdoc-args = ["--cfg", "docsrs"]
...
-hex-color = ["hex"]
+theme-from-image = ["kmeans_colors", "image"]
 
 [dependencies]
-palette = {version = "0.7", features = ["serializing"] }
-anyhow = "1.0"
-hex = {version = "0.4.3", optional = true}
+# palette = {version = "0.7", features = ["serializing"] }
+almost = "0.2"
+palette = {git = "https://github.com/Ogeon/palette", features = ["serializing"] }
```
ids1024, line 20: Palette 0.7.3 was just released. I presume that has what we need?

Cargo.toml
```diff
@@ -97,3 +107,7 @@ exclude = [
 
 [patch."https://github.com/pop-os/libcosmic"]
 libcosmic = { path = "./", features = ["wayland", "tokio", "a11y"]}
+
+# TODO Remove me when the palette crate gets an update & before merging
+[patch.crates-io]
+palette = {git = "https://github.com/Ogeon/palette", features = ["serializing"] }
```
ids1024, line 113: A `patch` section only works in a top-level crate, so if this is used as a dependency by something else (like cosmic-applets) the patch doesn't apply and it fails to build.

wash2: Oh, ya, that needs to be removed. It was removed from the theme update branch but not this one.

ids1024:
This actually wasn't compiling with the `applet` feature, so I added a commit that fixed two little things that were breaking it.

Using this in an applet, it seems like https://github.com/pop-os/libcosmic/commit/20a5227eca5d566427ecdfdb132ec9bd4becc2f1 has changed the styling for the applet buttons and buttons within them so the `hover` state isn't (very) visibly different. Changing `Button::Text => &cosmic.text_button` back to `Button::Text => &theme.current_container().component` reverts to the previous styling. But I guess that's something we want to change in the default theme instead?

ids1024:
Otherwise this is looking good. I wonder if `style()` could be handled automatically somehow when `applet::run` is used to avoid that (small) boilerplate. And then there's the need to create a `CosmicAppletHelper` while the code in `cosmic::app::applet` is already doing that, though that was already duplicated.

wash2:
> And then there's the need to create a `CosmicAppletHelper` while the code in `cosmic::app::applet` is already doing that, though that was already duplicated.

I think applets can just use `self.core.applet_helper` now instead of creating an `CosmicAppletHelper`, or am I missing something?

ids1024:
Ah right, it's a property of `core`. So yeah, that part is simple enough.

ids1024:
It doesn't seem quite right with the last commit either. The power applet here is styled using this branch, while the other applets are old versions.

eDP-1

Presumably when the applet isn't focused we still want the background to match the panel or be transparent.

wash2:
> It doesn't seem quite right with the last commit either. The power applet here is styled using this branch, while the other applets are old versions.

Ahh ok, I see what you mean. I'll just make it use a custom style then?

mmstick:
Perhaps the hover state visibility could be a property of the Appearance struct, or as an input parameter to our methods?

ids1024:
Looks like the appearance when not hovered is still like my last screenshot. With a somewhat darker circle for the button instead of matching the panel.

Previously this used a completely transparent background for the button when not hovered, and an opaque background when hovered. So I guess we need that to get the same behavior.

A over A = A for opaque colors, but for semi-transparent colors it gets the same color but with a higher alpha.

wash2:
> Looks like the appearance when not hovered is still like my last screenshot. With a somewhat darker circle for the button instead of matching the panel.
> 
> Previously this used a completely transparent background for the button when not hovered, and an opaque background when hovered. So I guess we need that to get the same behavior.
> 
> A over A = A for opaque colors, but for semi-transparent colors it gets the same color but with a higher alpha.

Oh sorry I forgot to make the buttons text buttons again after making the fix. Backgrounds of text buttons should be completely transparent.

ids1024 (approved):
I added a commit to not have "applet" depend on "a11y" since that was causing a panic... hopefully we can get that working well later.

Seems to be working fairly well now.

PR #140, opened 2023-08-10
Fix derive setter for default_icon_theme.

edfloreshz:
Uses manual implementation for setting the icon theme so that an empty string is treated as `None`.

Fixes this error message in the application example

```sh
error[E0277]: the trait bound `Option<String>: From<&str>` is not satisfied
  --> examples/application/src/main.rs:24:29
   |
24 |         .default_icon_theme("Pop")
   |          ------------------ ^^^^^^^^^ the trait `From<&str>` is not implemented for `Option<String>`
   |          |
   |          required by a bound introduced by this call
```

### DIFF
--- a/src/app/settings.rs
+++ b/src/app/settings.rs
@@ -26,7 +26,7 @@ pub struct Settings {
     pub(crate) default_font: Font,
 
     /// Name of the icon theme to search by default.
-    #[setters(into, strip_option)]
+    #[setters(skip)]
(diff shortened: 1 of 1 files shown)


mmstick (approved):
Seems fine to me, if it's okay with everyone else. An empty string doesn't allocate, and it's not a valid input anyway, so it makes sense to treat it as `None`.

mmstick (approved):

PR #142, opened 2023-08-15
update iced and add cosmic-time

wash2:


cosmic-time/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+documentation = "https://docs.rs/cosmic-time"
+keywords = ["gui", "animation", "interface", "widgets", "iced"]
+categories = ["gui"]
+
+[features]
+default = []
+once_cell = ["dep:once_cell"]
+
+[dependencies]
+libcosmic = { path = "../", default-features = false, features = [ "tokio" ] }
```
ids1024, line 18: So, `libcosmic` is a hard-dependency now, while it's optional in https://github.com/pop-os/cosmic-time/tree/libcosmic.

If we're adding cosmic-time to this repository, presumably we want to archive the `cosmic-time` repo and not have it in two places. And it would be nice if the library is still usable with upstream Iced? Though that's a bit awkward to support.

wash2: I was thinking about just keeping them as two separate repos for now, I guess I'll probably try to refactor it so it can use the shared stuff from the cosmic-time repo, but I think it'd be awkward to support both libcosmic and upstream iced in a single repo.

mmstick: We could make it a submodule, as we do for iced. I wonder what's necessary to merge it into cosmic-time with a feature flag.

wash2: > I wonder what's necessary to merge it into cosmic-time with a feature flag

Do you mean just using either iced 0.10.0 or libcosmic as dependencies inside cosmic-time, and also making it a submodule in libcosmic?

mmstick: Having cosmic-time in a single repository that supports iced and libcosmic with conditional compilation.

wash2: Ok if I understand you correctly, that is how it more or less was before this, but it wasn't really completed. I guess I could just do that and undo the changes here. It is just awkward when they both depend on each other, even though it's just the libcosmic examples that depend on cosmic-time.

mmstick (approved):

PR #143, opened 2023-08-17
Text input

wash2:
This is almost ready. Right now it supports all of the designed text inputs, except for the animated expandable text input. But I've currently only enabled it with the wayland feature. I am a bit unsure of how to handle text with multiple colors, for example, selected text taking on_accent, while regular text would be on_component , so for now it just does one color. I'm going to look more into how I can do that, but I'm not sure if that should block this.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -124,3 +124,5 @@ exclude = [
 
 [patch."https://github.com/pop-os/libcosmic"]
 libcosmic = { path = "./", features = ["wayland", "tokio", "a11y"]}
+[patch."https://github.com/pop-os/cosmic-time"]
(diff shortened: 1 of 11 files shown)


mmstick:
I'm fine with merging if the only concern is handling multi-color text. That's something we can add later. Is there anywhere where multi-color text input is used?

wash2:
> I'm fine with merging if the only concern is handling multi-color text. That's something we can add later. Is there anywhere where multi-color text input is used?

Not really that I can think of, aside from a text input with selected text, where the highlighted text should take the on_accent color instead of the regular text color

mmstick (approved):

PR #150, opened 2023-08-29
feat!: button & icon widgets rewrite

mmstick:
Closes https://github.com/pop-os/libcosmic/issues/154
Requires https://github.com/pop-os/iced/pull/59
Requires https://github.com/pop-os/cosmic-time/pull/16 

> There remains some smaller issues, but they'll be easier to address in future PRs so as to not delay the main improvements

This PR resurrects some of the improvements I had been working on earlier this year. For comparison, this adds a new [design demo example](https://github.com/pop-os/libcosmic/blob/buttons/examples/design/src/main.rs) that we'll use to showcase our widgets and their styles. I'll be migrating the widget demos from `cosmic` and `cosmic_sctk` here, and I'll work on getting it packaged so UX can actively follow the progress of our implementations.

To accommodate the needs of our new button designs; and to fix the poor ergonomics of the existing APIs; I've rewritten the button and icon widget APIs.  The new APIs are designed to be easier to configurable than before. Much less code will be required to create buttons and icons moving forward. It'll be significantly easier to configure them with chained methods, and there are now dedicated functions that we can use that serve as templates for preferred designs.

Here's a quick overview of the main changes:

```rs
use cosmic::widget::{button, icon};
use cosmic::theme;

/// An `icon::Handle` which may be an `Named` icon, `image::Handle`, or `svg::Handle`
/// This also has a configurable `symbolic` property to enable auto-coloring.
let firefox_handle = icon::from_name("firefox").size(16).handle();

// Suggested label button with leading and trailing icons
let suggested_button = button::suggested("Do something useful")
    .leading_icon(icon::from_name("bluetooth-active-symbolic"))
    .trailing_icon(firefox_handle.clone())
    .on_press(Message::ProductiveClick);

// Destructive label button with icon loaded by path
let destructive_button = button::destructive("Do something dangerous")
    .trailing_icon(icon::from_path("/custom/image.png"))
    .icon_size(16)
    .on_press_maybe(Some(Message::DestructiveClick));

// Label button that doesn't stand out
let standard_button = button::standard("Click me");

// Label button whose shape is invisible until hovered
let text_button = button::text("I'm clickable");

// Icon buttons may have optional labels, and a vertical orientation.
let icon_button = button::icon(icon::from_name("icon-name").size(24))
    .label("Optional label")
    .tooltip("This is shown on hover of the button")
    .vertical(true);
    
// Custom button with a manually-applied style
let custom = button(
    row()
        .spacing(8)
        .push(text("custom button"))
        .push(firefox_handle.clone().icon())
)
.style(theme::Button::Custom(appearance_fn));
```

Since these changes require refactoring most of libcosmic — and to ensure consistency in our designs and implementations in the future — all of the internal implementations now rely solely on importing widgets from libcosmic Any widgets from iced that we depend on have been re-exported to our widget module. This'll make it easy to adapt to any drop-in replacements in the future, and reduce any confusion about where to import widgets from.

This also requires an adjustment to `iced` o enable our `Theme` to have full control over the preferred default color of symbolic icons. We were previously applying styles to SVG widgets to color them, but we can now set this contextually so it'll no longer be necessary to attach styles to icons. The theme will set a preferred icon color application-wide in the `application::Appearance`, which can be overridden by a `container::Appearance`, or individually applied to specific `button::Appearance`s. The ability to set a custom SVG style option is still available if necessary.

Finally, I've added a prelude module that contains our essential extension traits and types. Iced recently started adding `_maybe()` methods to widgets, such as `on_press_maybe`, so I've started adding similar methods to our widgets. I've also added a `CollectionWidget` trait which adds a similar `.push_maybe()` method to columns and rows so we can chain widgets conditionally onto them, so that we needn't rely on macros or `Vec` to push widgets.

### Changes in iced

- [x] New icon_color renderer style for automatic coloring of symbolic icons
    - [x] Adds `symbolic` property was added to SVG widgets

### Changes in cosmic-time

- [x]  Updated to support libcosmic changes

### New design demo

- [x] Debian packaging
- [x] Debug page
    - [x] Choose between themes
    - [x] Toggle debug layout mode
    - [x] Change scaling factor
    - [ ] InfoBar
- [x] Buttons page
    - [x] Label Buttons
        - [x] With and without leading and trailing icons
        - [x] Disabled and enabled states
        - [x] Suggested
        - [x] Destructive
        - [x] Standard
        - [x] Text
    - [x] Icon Buttons
        - [x] With and without labels|
        - [x] Disabled and enabled states
        - [x] `extra_small` through `extra_large`
        - [x] Symbolic and non-symbolic icons
        - [x] Vertical orientation
        - [ ] Drop target style
    - [x] Link Buttons
- [x] Cards and ListViews page
    - [x] Cards
    - [ ] ListViews
- [x] Controls page
    - [x] Checkboxes
    - [ ] Control bar
    - [x] Radio buttons
    - [x] Spin buttons
    - [x] Sliders
    - [ ] Toggle buttons
    - [ ] Image buttons
    - [x] Progress bars
- [ ] Dividers page
- [ ] FlexRow page
- [ ] Navigation and Sidebars page
- [ ] Pickers page
- [x] Tabs and segmented controls page
- [x] Text input page
    - [x] Text Input
    - [x] Search Input
    - [x] Secure Input
    - [x] Inline Input

### Changes in cosmic-theme

- [x] Added `icon_button` component
- [x] Added `link_button` component
- [x] Fixed disabled component colors
- [x] Use of new state overlay colors based on dark/light status

### Changes in `cosmic::app`

- [x] Added `::command::set_theme()`
- [x] Added `::command::set_scaling_factor()`
- [x] Added `core.window.header_title`
    - So that the window title can differ from the header title
- [x] Removed theme property from `Core`
    - It was made redundant when the theme was set globally
    - The property didn't actually have any affect as a result

### Changes in `cosmic::theme`

- [x] Migrated all stylesheet implements to `::stylesheets`
    - [x] Removed defunct `expander` widget style
    - [x] Moved cosmic `TextInput` stylesheet impl here
    - [x] New cosmic `Button` stylesheet impl here
    - [x] Moved iced widget stylesheets to `::iced`
        - [x] Renamed iced `TextInput` to `IcedTextInput`
        - [x] Renamed iced `Button` to `IcedButton`
        - [x] Shared stylesheets re-exported to parent module
- [x] Renamed `theme()` to `system_preference()`
    - Renamed because it describes the purpose better
    - This function loads the preferred system theme from `cosmic-config`
- [x] Added `active()`, which returns the currently-active `Theme`
- [x] Added `active_type()` to return just the `ThemeType`
- [x] Added `is_dark()` and `is_high_contrast()`
- [x] Implemented `::is_dark()` and `::is_high_contrast()` for `ThemeType`
    
### Changes in `cosmic::widget`

- [x] Imports shared iced widgets to the widget module so we needn't import them from iced directly
    - [x] Refactored all internal implementations to use widgets from libcosmic
- [x] Makes text widget module public, and adds typography preset functions
    - title1, title2, title3, title4, heading, caption heading, caption, body, monotext
- [x] Adds column and row widget modules with new functions for creating them within
    - [x] `row()`, `row::with_capacity()`, `row::with_children()`
    - [x] `column()`, `column::with_capacity()`, `column::with_children()`
    - [x] Refactored all internal implementations for the new functions
- [x] Added tooltip module with tooltip container style
- [x] Adds `on_{{action}}_maybe()` methods that were missing from our widgets
- [x] Segmented button widget now accept `Icon` widgets directly, instead of `IconSource`
    - [x] There's now much less generic syntax because it's specialized to the cosmic theme now
- [ ] `Warning` is now the `InfoBar` widget
    - [ ] `::info()` to create info style
    - [ ] `::warning()` to create warning style
    - [ ] `::error()` to create error style
- [x] Applications, buttons, and containers now implement `icon_color` styles
    - [x] Text and symbolic icons may now have the same coloring
    - [x] Icon colors change based on dark and light them in regular buttons
    - [x] SVG styles removed as they're no longer necessary
- [x] Rewrites button widget internally to support cosmic features
    - [x] Adds styling parameters for configuring the outline width and color
    - [x] Adds styling parameter to configure symbolic icon coloring
    - [x] Modifies `StyleSheet` trait so our theme can match Figma designs
    - [x] Supports passing icon color from the renderer style if not applied to button
    - [x] New functions for creating suggested, destructive, standard, text, icon, and link buttons
    - [x] Changes necessary to `cosmic-theme` for the new button color changes
- [x] Rewrites button widget builder API with more flexible replacement
    - [x] New `Builder` type with branching type state paths
        - [x] TextButton builder
            - [x] may have leading and trailing icons
        - [x] IconButton builder
            - [x] come in many different sizes, from extra small to extra large
            - [x] optional label which may be horizontal or vertical
        - [x] LinkButton builder
            - [ ] May be used with new `cosmic::command::open_uri()` command
    - [x] Fixes display of buttons so that they meet our design specifications
    - [x] Much improved integration with the new icon APIs
    - [x] Tooltip integration
- [x] Rewrite of the icon and icon handle APIs
    - [x] `IconSource` removed, as the `Handle` now assumes the same role
    - [x] `Icon` widgets are clonable now because handles are Arc'd internally and cheap to clone
    - [x] New builder types and functions to load icons in different ways with less code
    - [x] Automatically tags icons as symbolic if imported by name when ending with `-symbolic`
    - [x] Provides a `symbolic()` method to manually set this state

### Smaller changes to libcosmic

- [ ] `cosmic` example replaced by `design-demo`
- [x] `cosmic-sctk` example replaced by `design-demo`
- [x] Builds no longer have default features
    - [x] Added compilation error if neither `wayland` or `winit is chosen
- [x] Adds monospace font to the default fonts: `FONT_MONO_REGULAR`
- [x] Adds `CollectionWidget` trait to extend column and row widget methods
    - [x] `::append()` to append multiple generic widgets from a `&mut Vec`
    - [x] `::extend()` to append multiple generic widgets from an iterator
    - [x] `::push_maybe()`, for chaining elements that are are conditional


src/app/cosmic.rs
```diff
@@ -274,13 +273,6 @@ impl<T: Application> Cosmic<T> {
             }
 
             Message::ThemeChange(theme) => {
-                // our system theme is always receiving updates so we should use it instead
```
wash2, line 277: I think something like the previous implementation with two different messages for theme changes would be better. Implemented this way, after setting the theme to custom in the example, making a change to the system theme overwrites the custom theme used by the application back to the system theme.

wash2 (commented):
looks great!

wash2 (approved):

PR #153, opened 2023-08-29
Fixed crate dependency issue

RachaelAva1024:
The module text_input was loaded at line 80, which conflicts with the crate text_input as declared on line 15. Cargo suggests binding the crate name as other_text_input, however, I named it crate_text_input to be more descriptive. I do not know if it will be an issue later in development, but libcosmic does at least compile successfully.


mmstick:
Fixed by #151

RachaelAva1024:
My apologies.

PR #162, opened 2023-09-18
Fix styling issues

edfloreshz:
This PR fixes some styling issues I missed on my last PRs.

### DIFF
--- a/src/applet/mod.rs
+++ b/src/applet/mod.rs
@@ -28,7 +28,7 @@ pub fn button_theme() -> Button {
     Button::Custom {
         active: Box::new(|active, t| widget::button::Appearance {
             border_radius: 0.0.into(),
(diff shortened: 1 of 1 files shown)


mmstick (approved):

src/applet/mod.rs
```diff
@@ -28,7 +28,7 @@ pub fn button_theme() -> Button {
     Button::Custom {
         active: Box::new(|active, t| widget::button::Appearance {
             border_radius: 0.0.into(),
-            ..t.active(active, &Button::Text)
+            ..t.active(active, &Button::Standard)
```
wash2, line 31: Why use the standard button style here instead of the Text button style? From what I've seen in the prototypes, the panel buttons should be transparent unless hovered.

PR #174, opened 2023-09-29
feat: color picker

wash2:
requires https://github.com/pop-os/iced/pull/64 or https://github.com/pop-os/iced/pull/63.

I'm looking into why the tool-tip copy buttons don't seem to be working, but otherwise I just need to fix the paths for iced and this is ready


mmstick:
The example should be a PR to the design demo, which can be done after merging this.

mmstick (approved):

PR #191, opened 2023-10-26
feat: enable limiting applications to a single instance

wash2:
This could allow applications to only allow a single instance by default, and when another instance is started, the arguments are passed to the existing application. I could see this possibly being useful for the applets that need to open a specific page of settings. They could spawn `cosmic-settings --network` for example, which might open a new instance on the network configuration page, or if there is one running, the arguments would be passed to the existing instance via dbus. Maybe we should make the interface match d-bus activation? This implementation is a bit different from GApplication and doesn't really use d-bus activation.


wash2:
I'll make a PR with dbus-activation instead

PR #215, opened 2023-11-19
Refactor

danieleades:
a little tidying up to squash some lint errors


mmstick (approved):

src/widget/text_input/input.rs
```diff
@@ -1515,18 +1515,6 @@ where
...
-            | wayland::DataSourceEvent::DndFinished
-            | wayland::DataSourceEvent::Cancelled,
-        ))) => {
-            let state = state();
-            if matches!(state.dragging_state, Some(DraggingState::Dnd(..))) {
-                state.dragging_state = None;
-                return event::Status::Captured;
-            }
-        }
-        #[cfg(feature = "wayland")]
```
mmstick, line 1529: What's the reason for removing this ?

danieleades: It has the same body as another match arm, and a more restrictive pattern, so is redundant

mmstick (approved):

PR #219, opened 2023-11-21
refactor: make the single-instance feature additive

wash2:
This is a refactor of the single instance / dbus activation implementation that makes the feature purely additive. It does add a generic parameter to Core, but I'm not sure of a better way of requiring it in `run_single_instance` while keeping `single-instance` feature additive.

### DIFF
--- a/examples/cosmic/Cargo.toml
+++ b/examples/cosmic/Cargo.toml
@@ -8,7 +8,7 @@ publish = false
 [dependencies]
 apply = "0.3.0"
 fraction = "0.13.0"
-libcosmic = { path = "../..", features = ["debug", "winit", "tokio"] }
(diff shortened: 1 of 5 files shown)


wash2:
https://github.com/pop-os/cosmic-settings/pull/102 has the updates for the settings app.

mmstick:
Would it be possible add a `DbusActivation` variant to `cosmic::app::cosmic::Message` for this? Then we could add a `dbus_activation` method to `cosmic::Application` for handling it.

wash2:
> Would it be possible to add a `DbusActivation` variant to `app::cosmic::Message` for this? Then we could add a `dbus_activation` method to `app::Application` for handling it.

Thanks, that would work well

mmstick (approved):

PR #226, opened 2023-11-29
fix: Reduce heading line height from 20 to 1

jacobgkau:
This fixes the settings drawer headings having too much vertical padding, reported at https://github.com/pop-os/cosmic-edit/issues/49.

Before:
Screenshot from 2023-11-28 22-24-26

After:
Screenshot from 2023-11-28 22-24-02

[Figma design](https://www.figma.com/file/s2o49ResysE3Q74jRwZrZV/COSMIC-Text-Editor?type=design&node-id=693-90896&mode=design&t=onjDJeoUEeXYzeHJ-0) for reference:
image

### DIFF
--- a/src/widget/text.rs
+++ b/src/widget/text.rs
@@ -50,7 +50,7 @@ pub fn title4<'a>(text: impl Into<Cow<'a, str>> + 'a) -> Text<'a, Renderer> {
 pub fn heading<'a>(text: impl Into<Cow<'a, str>> + 'a) -> Text<'a, Renderer> {
     Text::new(text)
         .size(14.0)
-        .line_height(20.0)
+        .line_height(1.0)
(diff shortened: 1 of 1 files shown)


jacobgkau:
This apparently isn't needed anymore after switching cosmic-edit off of the WGPU renderer in https://github.com/pop-os/cosmic-edit/commit/7b0d59785ca2e0e69c8506a03897ebfb1b8c0b60. I've confirmed cosmic-edit `7b0d597` displays correctly without this fix (actually, with the softbuffer renderer, the menu seems to look identical with this value set to 20.0 or 1.0.) Not entirely sure what's going on here, but I'll close this for now.

Ref:
https://github.com/pop-os/cosmic-edit/issues/49 
https://github.com/pop-os/libcosmic/issues/227

PR #228, opened 2023-11-30
WIP: Update to 0.12

wash2:
cosmic-time still needs to be updated. Most of libcosmic is working well now, but I'm still looking into a small layout issue for dropdown menu items, and also a color picker tooltip. It panics when interacting with the tooltip after interacting with the color picker previously.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -143,5 +143,3 @@ exclude = ["examples/design-demo", "iced"]
 [patch."https://github.com/pop-os/libcosmic"]
 libcosmic = { path = "./" }
 
-# [patch."https://github.com/pop-os/cosmic-time"]
-# cosmic-time = { path = "../cosmic-time" }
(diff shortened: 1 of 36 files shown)


wash2:
Everything seems to be working alright now, so this should be ready for review.
I will start updating some applications and applets to see if we need any more changes too.

https://github.com/pop-os/iced/pull/85
https://github.com/pop-os/cosmic-time/pull/17

wash2:
The launcher window is very small when testing.

mmstick:
Seems fine. There is one change that I noticed in the design demo. The width of the debug toggle, using cosmic-time's toggler, now defaults to `Length::Shrink` rather than `Length::Fill` now.

wash2:
> Seems fine. There is one change that I noticed in the design demo. The width of the debug toggle, using cosmic-time's toggler, now defaults to `Length::Shrink` rather than `Length::Fill` now.

Ok, should I change that? I had to change the method for the layout of the toggler in cosmic-time, which is probably what causes the difference.

mmstick:
We can adjust it in the design-demo

mmstick (approved):

PR #245, opened 2023-12-15
Optional Config dbus

wash2:


wash2:
This can be tested by running the multiwindow example and https://github.com/pop-os/cosmic-settings-daemon/pull/3 and then changing the theme

mmstick (approved):

PR #247, opened 2023-12-26
Remove hardcoded paths in cosmic-config

nbdd0121:
This commit changes the hardcoded /usr/share paths in cosmic-config to become performed via XDG lookups using the `xdg` crate.

This allows the installed files to be discovered on non-FHS Linux, e.g. NixOS.

Hardcoded /var/lib/ is removed entirely because
1. nothing installs to it yet (only user of new_state is cosmic_bg currently and it does not install to /var/lib)
2. it's intended for system states, not template for user state.
3. it's not part of XDG spec.

### DIFF
--- a/cosmic-config/Cargo.toml
+++ b/cosmic-config/Cargo.toml
@@ -10,7 +10,6 @@ macro = ["cosmic-config-derive"]
 subscription = ["iced_futures"]
 
 [dependencies]
-# For redox support
 zbus = { version = "3.14.1", default-features = false, optional = true }
 atomicwrites = { git = "https://github.com/jackpot51/rust-atomicwrites" }
 calloop = { version = "0.12.2", optional = true }
@@ -24,3 +23,9 @@ iced_futures = { path = "../iced/futures/", default-features = false, optional =
 once_cell = "1.19.0"
 cosmic-settings-daemon = { git = "https://github.com/pop-os/dbus-settings-bindings", branch = "cosmic-settings-daemon", optional = true }
 futures-util = { version = "0.3", optional = true }
+
+[target.'cfg(unix)'.dependencies]
+xdg = "2.1"
+
+[target.'cfg(windows)'.dependencies]
+known-folders = "1.1.0"
--- a/cosmic-config/src/lib.rs
+++ b/cosmic-config/src/lib.rs
@@ -94,8 +94,21 @@ pub trait ConfigSet {
 
 #[derive(Clone, Debug)]
 pub struct Config {
-    system_path: PathBuf,
+    system_path: Option<PathBuf>,
     user_path: PathBuf,
+}
+
+/// Check that the name is relative and doesn't contain . or ..
+fn sanitize_name(name: &str) -> Result<&Path, Error> {
+    let path = Path::new(name);
+    if path
+        .components()
+        .all(|x| matches!(x, std::path::Component::Normal(_)))
+    {
+        Ok(path)
+    } else {
+        Err(Error::InvalidName(name.to_owned()))
+    }
 }
 
 impl Config {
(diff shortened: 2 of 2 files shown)


nbdd0121:
If multiple OS support is in mind, it's very likely that libcosmic itself needs to handle the difference. On Linux XDG-based lookup is necessary, and this means that multiple directories need to be searched (at least for XDG_DATA_DIRS), so using the `dirs` for the `/usr/share` wouldn't be feasible since that crate gives only a single path.

nbdd0121:
How feasible would it be to support non-Unices BTW? I don't think Wayland would run on Windows at all (without wslg, that is)...

mmstick:
The dirs crate already handles this well.

nbdd0121:
xdg crate supports redox just fine. I am unaware that cosmic apps work on Windows though, could you elaborate how /usr/share is supposed to work on Windows?

I guess one approach is to use `dirs` crate for user paths, and `xdg` for system paths?

wash2:
> xdg crate supports redox just fine. I am unaware that cosmic apps work on Windows though, could you elaborate how /usr/share is supposed to work on Windows?
> 
> I guess one approach is to use `dirs` crate for user paths, and `xdg` for system paths?

Ya, or using `xdg` if the target os is linux or redox and `dirs` otherwise maybe

nyabinary:
What changed in the new commit?

nbdd0121:
> I guess one approach is to use `dirs` crate for user paths, and `xdg` for system paths?

This approach is used in the new commit, so the user path part is still cross-platform.

wash2:
The latest commit doesn't build on windows. I think avoiding the xdg crate when the target os is not redox or linux would probably work though.

nbdd0121:
I am not sure what the behaviour should be for Windows. Surely /usr/share/ also doesn't work?

Should I just set system_path to None on Windows?

wash2:
`%PROGRAMDATA%` seems to be a good option on windows from what I can tell.

nbdd0121:
I don't think ProgramData is the right analogy... It's supposed to be a location where shared user-immutable data files are placed, it feels like it should be where the programs are installed...

nyabinary:
> I don't think ProgramData is the right analogy... It's supposed to be a location where shared user-immutable data files are placed, it feels like it should be where the programs are installed...

Maybe doing programdata for now is good until a more permanent solution can be found?

mmstick:
Thoughts on using `"%SystemDrive%\Program Files\Common Files"` for Windows?

nbdd0121:
`%ProgramFiles%\Common Files\COSMIC` sounds a good location, although I think `Common Files` directory hasn't really be used by anyone lately.

Here're example paths that other programs use for shared stuff:
* LLVM: `%ProgramFiles%\share`
* Git: `%ProgramFiles%\usr\share`
* VS: `%ProgramFiles%\Microsoft Visual Studio\Shared`

I guess `%ProgramFiles%\COSMIC\{Common, Shared, share}` could also be suitable location?

mmstick:
Use whichever would be considered best practice for Windows. We can always change it later if necessary.

nyabinary:
@nbdd0121 any updates? :3

wash2 (approved):
This seems to be working well, thanks!

cosmic-config/Cargo.toml
```diff
@@ -24,3 +24,9 @@ iced_futures = { path = "../iced/futures/", default-features = false, optional =
 once_cell = "1.19.0"
 cosmic-settings-daemon = { git = "https://github.com/pop-os/dbus-settings-bindings", branch = "cosmic-settings-daemon", optional = true }
 futures-util = { version = "0.3", optional = true }
+
+[target.'cfg(any(unix, target_os = "redox"))'.dependencies]
```
jackpot51, line 28: redox is part of cfg(unix), so this can be simplified.

iced
jackpot51, line 1: This submodule change should be removed from this PR

mmstick:
I'll merge this to a dev branch and fix up the iced conflict.

cosmic-config/Cargo.toml
nbdd0121: Sorry, I missed this review. Fixed now.

mmstick (approved):

cosmic-config/Cargo.toml
```diff
@@ -24,3 +23,9 @@ iced_futures = { path = "../iced/futures/", default-features = false, optional =
 once_cell = "1.19.0"
 cosmic-settings-daemon = { git = "https://github.com/pop-os/dbus-settings-bindings", branch = "cosmic-settings-daemon", optional = true }
 futures-util = { version = "0.3", optional = true }
+
+[target.'cfg(any(unix))'.dependencies]
```
nbdd0121, line 27: Should be `cfg(unix)`

mmstick (approved):

PR #257, opened 2024-01-07
widget::popover: context menu positioning logic

jackpot51:
This keeps the popup inside the parent widget bounds, logic that is common for context menus


jackpot51:
As this is a minimal change that only has an effect on the COSMIC apps I am working on (ones using popover positioning), I will merge now.

PR #281, opened 2024-01-25
libcosmic: Add desktop-file helpers

Drakulix:
This is an attempt at unifying implementations of this across `cosmic-applibrary`, `cosmic-launcher`, `cosmic-app-list` (in cosmic-applets) and soon `cosmic-store`.

This is mostly based on the app-list implementation, as that parses most of the desktop files contents. Particular attention was given to allow for all the different filtering these apps do, as well as to easy icon handling. Additionally contains code necessary for handling gpu selection in the future (`prefers_dgpu` and explicitly setting a `wayland_display` for the spawn method).

This is just doing code deduplication and cleanup, the goal is not e.g. to do complete `exec` parsing in this PR.

I'll keep this as a draft until I have written proper implementations against all our use-cases, but in principle this is ready for review.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,8 @@ pipewire = ["ashpd?/pipewire"]
 process = ["dep:nix"]
 # Use rfd for file dialogs
 rfd = ["dep:rfd"]
+# Enables desktop files helpers
+desktop = ["process", "dep:freedesktop-desktop-entry", "dep:shlex"]
 # Enables keycode serialization
 serde-keycode = ["iced_core/serde"]
 # Prevents multiple separate process instances.
@@ -79,6 +81,8 @@ zbus = {version = "3.14.1", default-features = false, optional = true}
 
 [target.'cfg(unix)'.dependencies]
 freedesktop-icons = "0.2.5"
+freedesktop-desktop-entry = { version = "0.5.0", optional = true }
+shlex = { version = "1.3.0", optional = true }
 
 [dependencies.cosmic-theme]
 path = "cosmic-theme"
@@ -107,7 +111,6 @@ path = "./iced/futures"
 
 [dependencies.iced_accessibility]
 path = "./iced/accessibility"
-
 optional = true
 
 [dependencies.iced_tiny_skia]
--- a/src/desktop.rs
+++ b/src/desktop.rs
@@ -1,0 +1,213 @@
+pub use freedesktop_desktop_entry::DesktopEntry;
+use std::{
+    borrow::Cow,
+    ffi::OsStr,
+    path::{Path, PathBuf},
+};
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum IconSource {
+    Name(String),
+    Path(PathBuf),
+}
+
+impl IconSource {
+    pub fn from_unknown(icon: &str) -> Self {
+        let icon_path = Path::new(icon);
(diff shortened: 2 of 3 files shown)


mmstick:
I noticed that COSMIC applications weren't running on integrated graphics by default unless explicitly setting `WGPU_POWER_PREF=low`. It's currently being worked around by setting `low` on application startup from libcosmic. But we could use the `prefers_dgpu` property here to automatically set this environment variable `low` for integrated, and `high` for dedicated.

Drakulix:
> I noticed that COSMIC applications weren't running on integrated graphics by default unless explicitly setting `WGPU_POWER_PREF=low`. It's currently being worked around by setting `low` on application startup from libcosmic. But we could use the `prefers_dgpu` property here to automatically set this environment variable `low` for integrated, and `high` for dedicated.

So an app would lookup it's own desktop file and assuming it finds one check if it has the dGPU preference set? Sounds neat, but I don't think this would work in a sandbox like flatpak. I am honestly a bit annoyed of this behavior by wgpu. It should select whatever mesa picks by default to allow environment variables to work correctly.

jackpot51:
There's only two main env vars that will be set by switccheroo control, one for nvidia, one for amd. Our apps should set wgpu power pref if those are not set

jackpot51:
DRI_PRIME

the default GPU is the one used by Wayland/Xorg or the one connected to a display. This variable allows to select a different GPU. It applies to OpenGL and Vulkan (in this case “select” means the GPU will be first in the reported physical devices list). The supported syntaxes are:

    DRI_PRIME=N: selects the Nth non-default GPU (N > 0).

    DRI_PRIME=pci-0000_02_00_0: selects the GPU connected to this PCIe bus

    DRI_PRIME=vendor_id:device_id: selects the first GPU matching these ids.

For Vulkan it’s possible to append !, in which case only the selected GPU will be exposed to the application (eg: DRI_PRIME=1!).

Drakulix:
> There's only two main env vars that will be set by switccheroo control, one for nvidia, one for amd. Our apps should set wgpu power pref if those are not set

Honestly, parsing these environment variables manually feels like a hack as well and get more complicated when you consider things like `__NV_PRIME_RENDER_OFFLOAD_PROVIDER` to select a specific nvidia gpu.

I think the first thing we should do is neither force this to `low` or `high`, but use `WGPU_ADAPTER_NAME` to select a specific gpu. And for what gpu to select we should take the one advertised by the compositor, which we can get with a throw-away wayland/x11 connection.

Overall I am not happy we need to fix WGPU's enumeration here at all, but at least there is an open issue for that, so I am not entirely unhopeful that this will be fixed long-term upstream: https://github.com/gfx-rs/wgpu/issues/3464

(Possibly something we could contribute.)

jackpot51:
This looks good to me, it just needs a cargo fmt, anything else you want to do in this PR?

Drakulix:
Ready for review, now that I have verified, that this is good enough for all potential users:
- cosmic-app-list: Part of https://github.com/pop-os/cosmic-applets/pull/206
- cosmic-launcher: https://github.com/pop-os/cosmic-launcher/pull/84
- cosmic-app-library: https://github.com/pop-os/cosmic-applibrary/pull/76

mmstick (approved):

jackpot51:
Gonna leave the branch for you to delete when the prs are updated

PR #287, opened 2024-01-31
wip: update to use latest iced

wash2:
cosmic-time still needs to be updated, and then I need to test apps as well :sweat_smile:


src/app/cosmic.rs
```diff
@@ -65,6 +64,7 @@ pub enum Message {
     /// Capabilities the window manager supports
     #[cfg(feature = "wayland")]
     WmCapabilities(window::Id, WindowManagerCapabilities),
+    SurfaceClosed(window::Id),
```
mmstick, line 67: Needs comment here. Everything else looks good.

mmstick (changes requested):

mmstick (approved):

PR #379, opened 2024-04-06
add icon_button_with_handle for applet icon

### LINKED ISSUE #369: self.core.applet.icon_button should maybe take an handle instead of a str
I would like to use a custom icon

wiiznokes:
fix   #369

### DIFF
--- a/src/applet/mod.rs
+++ b/src/applet/mod.rs
@@ -12,7 +12,7 @@ use crate::{
     },
     iced_style, iced_widget,
     theme::{self, Button, THEME},
-    widget, Application, Element, Renderer,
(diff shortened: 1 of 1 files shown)


Drakulix:
Your applet would need a desktop file installed anyway to be picked up by the panel. Why do you want to set the icon programmatic instead of simply installing an icon file, like any other app would do?

wiiznokes:
> Your applet would need a desktop file installed anyway to be picked up by the panel. Why do you want to set the icon programmatic instead of simply installing an icon file, like any other app would do?

Hum, you right, idk why i didn't thought about it

PR #384, opened 2024-04-09
set config field take optional Config

wiiznokes:


cosmic-config-derive/src/lib.rs
```diff
@@ -89,11 +89,25 @@ fn impl_cosmic_config_entry_macro(ast: &syn::DeriveInput) -> TokenStream {
...
+                            }
+                        }
+                        None => {
+                            tracing::error!(
+                                "failed to save config {}: no config handler",
+                                stringify!(#field_name),
+                            );
+                        }
+                    }
                     Ok(true)
```
mmstick, line 111: This condition would return `Ok(true)` even though it failed to save the config.

wiiznokes:
> What is the use case for this?

The idea is that looking at app, they use a macro to set config fields, with this code because getting the `Config` can fail.

mmstick:
You can choose to ignore and log errors in your application. Same as what's being done in this PR.

wiiznokes:
> You can choose to ignore and log errors in your application. Same as what's being done in this PR.

But if i want to use the generated setters of my struct, i need to differentiate if the Config was successfully created or not. I think most application developer will not care if the config was set or not, since it's not recoverable.

I know it's possible to do this with a macro, but i tend to prefer rust code for autocompletion and readability

mmstick:
If you don't want to differentiate, then you can ignore the returned value.

```rs
let _result = config.set_property();
```

wiiznokes:
I meant if the `Config` was none, you can't call `config.set_bool(&config, false)`.

Or maybe it's ok to unwrap the config when the app is initialized, but it seems scrary

mmstick:
You would be better off just manually checking.

```rs
if let Some(ref config) = config {
   let _res =  config.set_bool(config, false);
}
```

wiiznokes:
> You would be better off just manually checking.
> 
> ```rust
> if let Some(ref config) = self.config {
>    let _res =  config.set_bool(config, false);
> }
> ```

It seems very boilerplate to me. And you don't handle the case of config being None so need to fallback to just changing the field: `config.bool = false`

I think the api of `cosmic_config::Config::new` should wrap the case of error, maybe by giving `Config::get_entry` a closure to generate the config in case of error, or just a Default bound. So that there is no a total api breakage and still requiring 1 line to set a config field. Wdyt ?

mmstick:
Your application should have already stored the config in your application's struct. So you only need to check if the stored config is `None`, as I did above.

wiiznokes:
> Your application should have already stored the config in your application's struct. So you only need to check if the stored config is None, as I did above.

As i said, this add boilerplate

```rust
if let Some(ref config) = self.config {
   let _res =  config.set_bool(config, false);
} else {
    config.bool = false;
}
```

instead of just `config.set_bool(config, false);`.

jackpot51:
I believe the proper place to handle this is in your application. I'd recommend just unwrapping the config handler early on, the only reasons for failure are filesystem issues where you'd probably not want to continue.

This change would require changes of all other programs using these functions to match. It would encourage applications to not handle their own error handling. It is not idiomatic.

wiiznokes:
I'm gonna stick with the macro, it's fine anyway. Thank you

PR #387, opened 2024-04-09
Segmented button Dnd

wash2:
This adds dnd support for segmented buttons and the nav bar. It can be tested with https://github.com/pop-os/cosmic-files/pull/103 and it should probably not be merged until https://github.com/pop-os/iced/pull/126

### DIFF
--- a/src/widget/nav_bar.rs
+++ b/src/widget/nav_bar.rs
@@ -7,12 +7,15 @@
 
 use apply::Apply;
 use iced::{
+    clipboard::{dnd::DndAction, mime::AllowedMimeTypes},
     widget::{container, scrollable},
     Background, Length,
 };
 use iced_core::{Border, Color, Shadow};
 
 use crate::{theme, widget::segmented_button, Theme};
+
+use super::dnd_destination::DragId;
 
 pub type Id = segmented_button::Entity;
 pub type Model = segmented_button::SingleSelectModel;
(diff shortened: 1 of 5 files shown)


wash2:
I think I'd like to remove the dnd id argument for the `nav_bar_dnd` method eventually and just rely on the widget id of the nav bar to not change, but this should work for now.

src/widget/nav_bar.rs
```diff
@@ -46,6 +49,44 @@ where
...
 
+/// Navigation side panel for switching between views.
+/// Can receive drag and drop events.
+pub fn nav_bar_dnd<Message, D: AllowedMimeTypes>(
+    model: &segmented_button::SingleSelectModel,
+    on_activate: fn(segmented_button::Entity) -> Message,
+    on_dnd_enter: impl Fn(segmented_button::Entity, Vec<String>) -> Message + 'static,
+    on_dnd_leave: impl Fn(segmented_button::Entity) -> Message + 'static,
+    on_dnd_drop: impl Fn(segmented_button::Entity, Option<D>, DndAction) -> Message + 'static,
+    id: DragId,
```
mmstick, line 60: I think we can add these to the `nav_bar` function. I'm not sure there's a good reason for us to not provide it by default.

wash2: I think that it's just a bit tricky to do because libcosmic also creates the nav bar internally, and then it would have to decide which message to produce, or maybe panic in the default dnd event handlers that should never be called if used with a type that has no allowed mime types. I'm not sure that it is useful to provide such methods though without knowing what type the user wants to accept.

mmstick (approved):

PR #409, opened 2024-04-24
improv(theme): theme generation improvements

mmstick:
This improves our theming by a decent margin

- Fixes background layer hover styling on the light theme
- Text color is now based on the lightness value of the container
- Buttons, SegmentedButtons, and the NavBar now inherit the text color as well
- Fixed styling of active buttons in segmented button controls
- The `@small_container_widget` variable is directly affected by container layers, so it's been promoted as a property of the `Container` struct. This is given a 0.25 alpha to match the Figma designs.

These are the main issues remaining

- Accent color has poor contrast when the container background is dark while the application background is light; and vice versa
- Neutral palette colors are reversed when the container background is dark while the application background is light; and vice versa
- Window stack header text color has poor contrast if a light mode theme has a dark application background; and vice versa
- Segmented button widget styling is still slightly off from Figma.

### DIFF
--- a/cosmic-theme/src/model/derivation.rs
+++ b/cosmic-theme/src/model/derivation.rs
@@ -15,18 +15,28 @@ pub struct Container {
     pub divider: Srgba,
     /// the color of text in the container
     pub on: Srgba,
+    /// the color of @small_widget_container
+    pub small_widget: Srgba,
 }
 
 impl Container {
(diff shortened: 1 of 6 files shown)


TheESN:
Also I want to point out that window stack header doesn't actually use base colour from theme

wash2 (approved):

mmstick:
This is now merged into the latest commit of cosmic-settings

PR #412, opened 2024-04-24
[feat] set default color palette based on user preference

elevenhsoft:
Hi

Today tried libcosmic on Pop_OS with current Gnome implementation. The problem was with preferred color palette in libcosmic apps. I'm using light theme in my OS and apps was started always as dark theme. Before I was added these feature to my app when using Iced and it was working so considered to add this to libcosmic. I'm not sure if you want to see such features. The implementation is very simple based on `gsettings` app and `color-scheme` value. From my tests on pop os it's working. It should work everywhere where gsettings are installed, otherwise it will default to dark theme as before so nothing lost here :) Would be nice to see it implemented.

### DIFF
--- a/cosmic-theme/src/model/theme.rs
+++ b/cosmic-theme/src/model/theme.rs
@@ -101,7 +101,7 @@ pub struct Theme {
 
 impl Default for Theme {
     fn default() -> Self {
-        Self::dark_default()
+        Self::preferred_theme()
     }
 }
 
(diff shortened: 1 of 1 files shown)


cosmic-theme/src/model/theme.rs
```diff
@@ -513,6 +513,40 @@ impl Theme {
...
+
+        Self::dark_default()
+    }
+
+    /// check current desktop environment and preferred color scheme and set it as default
+    pub fn preferred_theme() -> Self {
+        let current_desktop = std::env::var("XDG_CURRENT_DESKTOP");
+
+        if let Ok(desktop) = current_desktop {
+            if desktop.trim().to_lowercase().contains("gnome") {
```
bbb651, line 543: This value would never be anything other than exactly `GNOME` on GNOME, this seems a bit error prone and unnecessary, the best list I could find of possible values in on [this page](https://wiki.archlinux.org/title/Xdg-utils) on the arch wiki. This code only runs when `xdg-portal` is disabled right?

bbb651: I guess it does match `GNOME-Flashback` which it probably should, I've never heard of it before, I guess it's very unlikely a desktop name would contain "gnome", but with other desktops like "kde" it's more likely to have collisions

elevenhsoft: yeah maybe you are right. However pop-os return different value:

```
~ ❯ echo $XDG_CURRENT_DESKTOP
pop:GNOME
```
so it's in case some distros have this value set to something else than `GNOME`

PR #413, opened 2024-04-26
Fix icons and text in header bar not inheriting accent color from header bar container style

mmstick:
screenshot-2024-04-26-13-41-46


wash2:
Maybe I'm looking at out of date designs, but it seems like the headerbar text shouldn't be the accent color according to the page in figma.

wash2 (dismissed):

PR #426, opened 2024-05-06
fix: avoid setting WGPU_POWER_PREF=low if system is desktop

mmstick:
May fix desktops with integrated graphics having a delay on startup.


wash2 (approved):

Drakulix:
At this point, I wonder if we should just check if we connect via Wayland, then do a quick connection ourselves and get the primary device from wp-dmabuf-linux.

We are just doing more and more guesswork to estimate, what the compositor is most likely doing.

PR #436, opened 2024-05-13
Export vscode

wash2:
exports the theme to vs code. Some colors are still missing (borders and outlines mostly), or maybe could be changed to something more appropriate, but it looks pretty good to me. Can be tested with https://github.com/pop-os/cosmic-settings-daemon/pull/21


mmstick (approved):

wash2:
Hmm, I guess it should also probably apply `  "window.autoDetectColorScheme": true` so that other colors that aren't set look ok

PR #444, opened 2024-05-14
Update iced and drop wgpu_power_pref

Drakulix:
for testing, see https://github.com/pop-os/iced/pull/138


Drakulix:
> Will this work for the winit and x11 environments?

Winit on other systems will do whatever wgpu is doing there by default.

X11 currently would select the default as well, but could be easily extended to query DRI3 instead. X11 and multi-gpu is touchy in general, I wouldn't try to override this with WGPU_POWER_PREF, but let it do it's thing, but we can figure out the primary GPU via DRI3, if that is desired.

Drakulix:
Marking as ready for review. Now that https://github.com/pop-os/iced/pull/138 is merged, we can just drop the POWER_PREF.

mmstick (approved):

PR #449, opened 2024-05-16
feat: add validate_input text widget

elevenhsoft:
Hello,

I just got a though it would be nice addition to have some form of validation for inputs. It would indicate by border color if it meets condition or no. Something like web forms doing it.

For now, in my app I'm using awesome warning widget. IMHO it's not that good for my use case, since it takes much space and I'm not hiding it. Probably I would fix this issue in my app, but why not trying to push something to upstream? ^^

Added also simple window with 2 widgets, `text_input` and `validate_input` as a example. So you can test how it works.

### DIFF
--- a/examples/input_validation/src/main.rs
+++ b/examples/input_validation/src/main.rs
@@ -1,0 +1,83 @@
+use cosmic::app::{Application, Command, Core, Settings};
+use cosmic::iced;
+use cosmic::iced::executor;
+use cosmic::iced_core::Size;
+use cosmic::widget::{column, text_input, validate_input};
+use cosmic::Element;
+use regex::Regex;
+
+#[derive(Debug, Clone)]
+pub enum Message {
+    InputName(String),
+    InputEmail(String),
+}
+
+struct Window {
+    core: Core,
+    name: String,
+    email: String,
+}
+
+impl Application for Window {
+    type Executor = executor::Default;
+    type Flags = ();
+    type Message = Message;
+    const APP_ID: &'static str = "com.system76.InputValidation";
+
+    fn core(&self) -> &Core {
+        &self.core
+    }
+
+    fn core_mut(&mut self) -> &mut Core {
+        &mut self.core
+    }
+
+    fn init(core: Core, _flags: Self::Flags) -> (Self, Command<Message>) {
+        (
+            Self {
+                core,
+                name: String::new(),
+                email: String::new(),
+            },
+            Command::none(),
+        )
+    }
+
+    fn update(
+        &mut self,
+        message: Self::Message,
+    ) -> iced::Command<cosmic::app::Message<Self::Message>> {
+        match message {
+            Message::InputName(name) => self.name = name,
+            Message::InputEmail(email) => self.email = email,
+        }
+        Command::none()
+    }
+
+    fn view(&self) -> Element<Self::Message> {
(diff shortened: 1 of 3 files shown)


mmstick:
Make sure to run `cargo fmt` and set up `cargo clippy`

elevenhsoft:
Yes, sorry.

examples/input_validation/src/main.rs
```diff
@@ -0,0 +1,83 @@
...
+                email: String::new(),
+            },
+            Command::none(),
+        )
+    }
+
+    fn update(
+        &mut self,
+        message: Self::Message,
+    ) -> iced::Command<cosmic::app::Message<Self::Message>> {
```
mmstick, line 49: This should be `cosmic::app::Command<Self::Message>`

src/widget/text_input/input.rs
```diff
@@ -2028,6 +2053,12 @@ pub fn draw<'a, Message>(
         text_color = text_color.blend_alpha(background, 0.5);
     }
 
+    if is_validated.is_some() {
+        if !is_validated.unwrap() {
+            border_color = Color::from_rgb(0.95, 0.30, 0.30);
+        }
+    }
```
mmstick, line 2060:
```rs
if is_validated.unwrap_or_default() { }
```

src/widget/text_input/input.rs
```diff
@@ -160,6 +160,20 @@ where
...
+pub fn validate_input<'a, Message>(
+    placeholder: impl Into<Cow<'a, str>>,
+    value: impl Into<Cow<'a, str>>,
+    validate_condition: bool,
+) -> TextInput<'a, Message>
+where
+    Message: Clone + 'static,
+{
+    TextInput::new(placeholder, value).validation_condition(validate_condition)
+}
```
mmstick, line 175: Seems like this should be a property rather than its own separate widget

elevenhsoft: So we should use `text_input` and then add a property `validation_condition` for example and done?

src/widget/text_input/input.rs
```diff
@@ -2028,6 +2053,12 @@ pub fn draw<'a, Message>(
         text_color = text_color.blend_alpha(background, 0.5);
     }
 
+    if is_validated.is_some() {
+        if !is_validated.unwrap() {
+            border_color = Color::from_rgb(0.95, 0.30, 0.30);
```
mmstick, line 2058: This should get a color from the cosmic palette. Not sure what though

src/widget/text_input/input.rs
mmstick: Yes

elevenhsoft: Oh yeah. Got some fixes and removed widget. Just need to change palette and I will push it ^^

elevenhsoft:
Ok. I did some changes, meanwhile I did it again, did not cargo fmt. Sorry, it's late ^^

This time we also using color from `destructive` color palette, I think it should do job here. Unfortunately I was not sure how to use it dynamically so it's just as a absolute value. Is it done properly? I don't know but it do job just fine :)

src/widget/text_input/input.rs
elevenhsoft: Yeah, I just used a value from `destructive` color.

PR #452, opened 2024-05-17
Fix header bar buttons not inheriting icon and text color from headerbar container

mmstick:
This also requires specializing our button widget so it is now a `Button<'a, Message>` instead of `Button<'a, Message, crate::Theme, crate::Renderer>`.

screenshot-2024-05-17-17-11-50

### DIFF
--- a/src/app/core.rs
+++ b/src/app/core.rs
@@ -147,7 +147,7 @@ impl Default for Core {
                 height: 0,
                 width: 0,
             },
-            focused_window: None,
+            focused_window: Some(window::Id::MAIN),
             #[cfg(feature = "applet")]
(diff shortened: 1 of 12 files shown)


wash2 (changes requested):
It seems like the applet module also needs to be updated

wash2 (approved):

mmstick:
It seems that the window does not begin with a focus state. I wonder where that can be obtained

mmstick:
Active header bar state

screenshot-2024-05-17-18-36-28

Inactive header bar state

screenshot-2024-05-17-18-36-44

wash2 (approved):

PR #458, opened 2024-05-21
Add guide book for libcosmic app development

mmstick:
Work in progress. There is a challenge for publishing both a book and cargo docs.


mmstick:
Provided by https://pop-os.github.io/libcosmic-book/introduction.html

PR #471, opened 2024-05-27
Add header_size density config to configure header bar compactness

mmstick:
Can be tested with `echo Compact > ~/.config/cosmic/com.system76.CosmicTk/v1/header_size`
Then running `just run application`

This will also make the `CosmicTk` config a thread-local static global, which will make it easy to implement the interface_density config.


ryanabx:
Will this help with https://github.com/pop-os/cosmic-comp/issues/502

mmstick:
@ryanabx cosmic-comp could use `headerbar().density(Density::Compact)`

jackpot51 (approved):

PR #500, opened 2024-06-15
add markdown widget

elevenhsoft:
Hello,

This PR extends text module with new markdown widget. Everything is hidden behind `markdown` feature which needs to be enabled to work with this widget. Access is from `widget::text::markdown()` via helper function. This function require passing text in markdown format as a `String`. The markdown `parser` and `Markdown` widget are in different files because it's easier to work with it. From my testing it works pretty good, the text is formatted. There are some problems with links and images, so I made them as a text with URL (not clickable ofc), the same with image.

The second problem and I don't know how to solve it, is with strings which comes from `markdown` crate. For spans in `rich_text()` we need string slices (or no?) so I used `Box::leak()` to achieve this. Maybe someone have better idea, or ideal would be if spans can contains `Strings`. But I did not test it yet.

### DIFF
--- a/examples/markdown-widget/Cargo.toml
+++ b/examples/markdown-widget/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "markdown-widget"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+
+[dependencies.libcosmic]
+path = "../../"
+default-features = false
+features = ["winit", "wgpu", "markdown"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,14 @@ clipboard = ["iced_sctk?/clipboard"]
 dbus-config = ["cosmic-config/dbus", "dep:zbus", "cosmic-settings-daemon"]
 # Debug features
 debug = ["iced/debug"]
+# Enables markdown widget
+markdown = [
+    "dep:cosmic-text",
+    "dep:markdown",
+    "dep:syntect",
+    "dep:two-face",
+    "dep:cosmic-syntax-theme"
+]
 # Enables pipewire support in ashpd, if ashpd is enabled
 pipewire = ["ashpd?/pipewire"]
 # Enables process spawning helper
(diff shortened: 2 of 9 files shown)


examples/markdown-widget/Cargo.toml
mmstick, line 1: This example should be called `markdown`

wiiznokes:
Maybe this could be added directly in Iced, this is on the Iced roadmap: https://whimsical.com/roadmap-iced-7vhq6R35Lp3TmYH4WeYwLM

examples/markdown-widget/Cargo.toml
elevenhsoft: Well, that's what I did before this rename. However I don't know what's happened but it did not work with `markdown` crate, there was a dependency issue. I'm not sure but renaming this example fixed this issue :sweat_smile:

mmstick:
I'll check the string problem

elevenhsoft:
> Maybe this could be added directly in Iced, this is on the Iced roadmap: https://whimsical.com/roadmap-iced-7vhq6R35Lp3TmYH4WeYwLM

Yeah, I'm here after this comment: https://github.com/elevenhsoft/cosmic-ext-applet-ollama/issues/1#issuecomment-2168355798 ^^

I did know about Iced roadmap. They need to update cosmic-text tho if I'm correct.

wiiznokes:
Putting it in Iced would be better, but it will probably be merged after 2 years :smile:, so nevermind my coment. I may use it in fan-control for an explanation screen so its nice to see this :)

Btw, are you using fedora ? I tried to package your applet 2 hours ago but i can't install Ollama.

elevenhsoft:
> Btw, are you using fedora ? I tried to package your applet 2 hours ago but i can't install Ollama.

No, only Pop_OS and openSUSE.

elevenhsoft:
Hey @mmstick , I made some changes to Parser, tried to remove most of the `Box::leak()` with success. Some of them are still there but should not cause issues I hope :)

mmstick:
I will try to eliminate the remaining leaks soon.

jackpot51 (changes requested):
This is a lot of code, and the code from cosmic-edit for rendering a Buffer is not ideal. I want to help optimize this before merging to ensure it renders using iced's Buffer rendering code so it can be GPU accelerated.

elevenhsoft:
Thank you! :)

wiiznokes:
In case anyone missed it, there is now a markdown widget in iced: https://github.com/iced-rs/iced/pull/2508

elevenhsoft:
> In case anyone missed it, there is now a markdown widget in iced: [iced-rs/iced#2508](https://github.com/iced-rs/iced/pull/2508)

Thank you!

In this case I think we should use Iced's markdown. This one is... well, far from being good. There is still a lot of work to be done.

PR #517, opened 2024-06-27
feat: toast widget

wiiznokes:
Attempt to implement https://github.com/pop-os/cosmic-files/issues/183

I implemented this feature in fan-control. You can see the change i needed to make in [this commit](https://github.com/wiiznokes/fan-control/compare/master...toats).

The duration chosen comes from [Android](https://stackoverflow.com/questions/7965135/what-is-the-duration-of-a-toast-length-long-and-length-short).

There is more just to match the Ui design, but the idea is there. I'm not sure i will implement it as i don't really know where to search for padding, colours, ect..

Capture d’écran du 2024-06-28 00-21-00

### DIFF
--- a/src/widget/toaster/mod.rs
+++ b/src/widget/toaster/mod.rs
@@ -3,47 +3,39 @@
 
 use std::time::Duration;
 
-use iced::{Command, Length};
-use iced_core::{Element};
 use crate::widget::container;
 use crate::widget::Column;
+use iced::{Command, Length};
+use iced_core::Element;
 use widget::Toaster;
 
 use crate::ext::CollectionWidget;
 
 use super::{button, row, text};
 
-
 mod widget;
 
-pub fn toaster<'a, Message: Clone + 'a, Theme, Renderer>(
+pub fn toaster<'a, Message>(
     toasts: &'a Toasts<Message>,
-    content: impl Into<Element<'a, Message, Theme, Renderer>>,
-) -> Element<'a, Message, Theme, Renderer>
+    content: impl Into<Element<'a, Message, crate::Theme, iced::Renderer>>,
+) -> Element<'a, Message, crate::Theme, iced::Renderer>
 where
-    Renderer: iced_core::Renderer,
+    Message: Clone + 'a,
 {
+    let make_toast = |toast: &'a Toast<Message>| {
+        let row = row().push(text(&toast.message)).push_maybe(
+            toast
+                .action
+                .as_ref()
+                .map(|t| button(text(t.0.clone())).on_press(t.1.clone())),
+        );
 
-
-    let make_toast = |toast: &Toast<Message>| {
-
-        let row = row()
-            .push(text(toast.message))
-            .push_maybe(toast.action.map(|t| button(text(t.0)).on_press(t.1)));
-
-        container(row)
-            .into()
-
+        container(row).into()
     };
 
+    let col = Column::with_children(toasts.toasts.iter().map(make_toast));
 
-    let col = Column::with_children(toasts.toasts.iter().map(|t| {
-        make_toast(t)
-    }));
-
-
-
-    Toaster::new(col.into(), content.into()).into()
+    Toaster::new(col.into(), content.into(), toasts.toasts.is_empty()).into()
 }
 
 pub enum ToastDuration {
--- a/src/widget/toaster/mod.rs
+++ b/src/widget/toaster/mod.rs
@@ -1,0 +1,106 @@
+// Copyright 2024 wiiznokes
+// SPDX-License-Identifier: MIT
+
+use std::time::Duration;
+
+use iced::{Command, Length};
+use iced_core::{Element};
+use crate::widget::container;
+use crate::widget::Column;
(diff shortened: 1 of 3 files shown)


wiiznokes:
Should this be moved to core ?

src/widget/toaster/mod.rs
```diff
@@ -0,0 +1,175 @@
...
+// SPDX-License-Identifier: MIT
+
+//! A widget that display toats.
+
+use std::collections::VecDeque;
+use std::time::Duration;
+
+use crate::widget::container;
+use crate::widget::Column;
+use iced::Command;
```
mmstick, line 11: Should be `cosmic::app::Command`

src/widget/toaster/mod.rs
```diff
@@ -0,0 +1,175 @@
...
+        let duration = toast.duration.duration();
+
+        self.toasts.push_back(toast);
+
+        Command::perform(
+            async move {
+                tokio::time::sleep(duration).await;
+            },
+            |_| crate::app::Message::App(message.into()),
+        )
```
mmstick, line 165:
```rs
crate::command::future(async move {
    tokio::time::sleep(duration).await;
    Message::from(message)
})
```

wiiznokes:
So, should this be moved to core, like the navbar? @mmstick

mmstick:
Probably

wiiznokes:
With the current impl, it requires Core to be generic over Message, is it ok?

mmstick:
That kind of state should be in the app struct

wiiznokes:
The only idea i have is adding this function in the Application trait fn `toast_message(&self, toats_id: u32) -> Message;` but it leaks to toasts_id implementation details, and it force the user to keep track of his toasts.

Can you think of a better solution ? Maybe it will be preferable to not include in core afterall

mmstick:
I think it's fine for the app to store the model of the toaster. Similar to what it already does for other widgets.

src/widget/toaster/mod.rs
wiiznokes: I had to write 
```rs
crate::command::future(async move {
            tokio::time::sleep(duration).await;
            crate::app::Message::App(Message::from(message))
        })
```

src/widget/toaster/widget.rs
```diff
@@ -0,0 +1,277 @@
...
+
+impl<'a, Message, Theme, Renderer> From<Toaster<'a, Message, Theme, Renderer>>
+    for Element<'a, Message, Theme, Renderer>
+where
+    Renderer: renderer::Renderer + 'a,
+    Theme: 'a,
+    Message: 'a,
+{
+    fn from(
+        toggler: Toaster<'a, Message, Theme, Renderer>,
```
mmstick, line 273: Should be `toaster`

src/widget/toaster/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+        let row = row()
+            .push(text(&toast.message))
+            .push_maybe(toast.action.as_ref().map(|action| {
+                button::standard(&action.description).on_press(action.message.clone())
+            }))
+            .push(button::standard("close").on_press(ToastMessage(toast.id).into()))
+            .align_items(iced::Alignment::Center);
+
+        container(row)
+            .padding(10)
```
mmstick, line 39: This should get a padding from the cosmic theme.

```rust
.padding(cosmic::theme::active().cosmic().space_xxs()) // Default 8
.padding(cosmic::theme::active().cosmic().space_xs()) // Default 12
```

src/widget/toaster/mod.rs
```diff
@@ -0,0 +1,173 @@
+// Copyright 2024 wiiznokes
+// SPDX-License-Identifier: MIT
+
+//! A widget that display toats.
```
mmstick, line 4: s/toats/toasts

src/widget/toaster/mod.rs
```diff
@@ -0,0 +1,173 @@
+// Copyright 2024 wiiznokes
+// SPDX-License-Identifier: MIT
```
mmstick, line 2: MPL-2.0 preferred here

src/widget/toaster/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+            .push(button::standard("close").on_press(ToastMessage(toast.id).into()))
+            .align_items(iced::Alignment::Center);
+
+        container(row)
+            .padding(10)
+            .style(crate::style::Container::Card)
+    };
+
+    let col =
+        Column::with_children(toasts.toasts.iter().rev().map(|t| make_toast(t).into())).spacing(5);
```
mmstick, line 44:
```rust
let col = toasts.toasts.iter()
    .rev()
    .map(make_toast)
    .fold(column::with_capacity(toasts.toasts.len()), |col, t| col.push(t.into())
    .spacing(cosmic::theme::active().cosmic().space_xxxs());

mmstick (approved):

mmstick:
Thanks!

PR #528, opened 2024-07-08
Fixed broken link in README.md to point to docs. Added code to add libcosmic as dependency for lib.

Tech-Priest-Of-Rust:
Fixed broken link in README.md to point to documentation. 
Added code to add libcosmic as dependency for library instead of just a workspace. 
This makes the README.md more complete.
Please let me know if you need any changes as I am very flexible with my time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,15 @@ default-features = false
 features = ["wayland", "tokio"]
 ```
 
+Alternatively, to create a new package after running `cargo new {{name_of_project}}`,
+edit the `Cargo.toml` within, and edit the dependencies section to look like this.
+
+```toml
+[dependencies]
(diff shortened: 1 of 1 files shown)


jacobgkau (approved):
Engineering can feel free to let us know if they want to repurpose https://github.com/pop-os/docs for this, but as of right now, the new link is definitely much more relevant.

I'm not certain on the wording being added before the new code example, but I'll let engineering comment on that.

Tech-Priest-Of-Rust:
@jacobgkau how would you suggest changing it? I am a new contributor so I would like to know more about the format of the text.

jackpot51 (approved):

jackpot51:
We may want to link to https://pop-os.github.io/libcosmic-book/ as well.

Issue #538, opened 2024-07-17
COSMIC overrides my entire VSCode settings on every login


hgaiser:
Everytime I log in, COSMIC overrides my entire `settings.json` file for VSCode. All the settings are gone and replaced with only a COSMIC theme. The intend of `cosmic-theme` is to only inject the theming part, but [this](https://github.com/pop-os/libcosmic/blob/master/cosmic-theme/src/output/vs_code.rs#L280) line does not work for VSCode `settings.json` files.

VSCode `settings.json` does not strictly follow the json syntax. It, for instance, also allows trailing comma's and comments. If this syntax is used, `serde_json` produces an error. Since my `settings.json` file includes trailing comma's, COSMIC throws away the entire file because it fails to parse it.

The issue with `serde_json`:

```rust
fn main() {
    let _settings: serde_json::Value = serde_json::from_str("// Some comment\n{\"foo\": \"bar\",}").unwrap();
}
```

Produces the error:
```
thread 'main' panicked at src/main.rs:2:101:
called `Result::unwrap()` on an `Err` value: Error("expected value", line: 1, column: 1)
```

Or simply:

```rust
fn main() {
    let _settings: serde_json::Value = serde_json::from_str("{\"foo\": \"bar\",}").unwrap();
}
```

Produces:

```
thread 'main' panicked at src/main.rs:2:84:
called `Result::unwrap()` on an `Err` value: Error("trailing comma", line: 1, column: 15)
```

I'm not sure if there is a crate that can parse this style of json correctly (see [this](https://github.com/serde-rs/json/issues/168) issue).

Regardless, I'm not a fan of forcibly setting the theme for an application like VSCode. My suggestion would be to add a theme to the VSCode extension marketplace and direct users there, if they wish to use it.

Drakulix:
If anything, we probably shouldn't override a file we can't parse.

Also in the meantime, cosmic is only doing this, if you enable the experimental "Apply this theme to GNOME apps" option.

hgaiser:
> If nothing else, we shouldn't override a file we can't parse.
> 
> Also in the meantime, cosmic is only doing this, if you enable the experimental "Apply this theme to GNOME apps" option.

Ah thanks, I can at least disable this behaviour for now :)

hgaiser:
By the way, this behavior still seems to happen when "Apply this theme to GNOME apps" was disabled.

Vedantpr0:
[...]

PR #539, opened 2024-07-17
fix: return error if vscode config fails to parse, instead of overwriting

wash2:
workaround for #538


mmstick (approved):

hgaiser:
I updated all COSMIC packages to the latest commit but this still seems to happen (cosmic-settings-daemon is at r68). Any idea what could cause this?

PR #548, opened 2024-07-25
Call systemd `StartTransientUnit` when starting app

ids1024:
This is needed for things like `xdg-desktop-portal` to get the app ID from a pid, in unsandboxed apps.

A couple things still should be fixed here, then `cosmic-launcher`, `cosmic-app-list`, and `cosmic-app-library` will need to be updated to use this.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,8 @@ jobs:
         - winit_wgpu
         - wayland
         - applet
-        - desktop
+        - desktop,smol
+        - desktop,tokio
     runs-on: ubuntu-22.04
     steps:
       - name: Checkout sources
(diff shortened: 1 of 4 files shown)


ids1024:
I guess the use of tokio in https://github.com/pop-os/libcosmic/pull/548/commits/fd0e54747976c383f02212a753a2b0bde06af6c5, although a clean solution, is problematic if the `tokio` feature of `libcosmic` isn't enabled...

ids1024:
With the changes here:

- `cosmic::process::spawn` and `cosmic::desktop::spawn_desktop_exec` are now async functions.
- The `process` and `desktop` features now depend on either `smol` or `tokio` features being enabled. But if we depend on `zbus`, it will fail to compile anyway without one of those. Presumably cosmic apps should always enable one.
- If the `tokio` feature is enabled, the futures will need to be called in the tokio executor, which should be like other things.

These things seem reasonable.

Drakulix (approved):
LGTM

PR #553, opened 2024-07-30
fix/feat: titlebar icons on other os

### LINKED ISSUE #523: Draw native window controls within the custom titlebar on non linux distributions
I think that would result in a better experience for cross platform support, without removing anything from the custom titlebar.

On windows, this allows for the snap tool to work by hovering over the minimize or maximize button.

On Mac os, forcing the window controls to be on the right whilst all controls are on the left feels very weird.

wiiznokes:
should fix #523


mmstick (approved):
There are a handful of other icons being pulled from the icon theme as well. May want to provide functions for each of the icons required.

PR #557, opened 2024-08-02
fix: Use globals instead of thread-locals

Drakulix:
Better support for multi-threaded applications, especially cosmic-comp rendering in parallel on multiple threads, each potentially accessing global configurations such as the active theme, icon_theme and more...

This is already causing issues like https://github.com/pop-os/cosmic-comp/issues/671

Note: Leaves the SUBSURFACE-thread-local untouched. It is not part of the public api and looking at it, I am not sure it is possible to have access to this from multiple threads. cosmic-comp also has no business with iced-sctk, so this a non-issue for the cases I want to fix here. (Ping @ids1024 on this specifically.)

### DIFF
--- a/src/app/cosmic.rs
+++ b/src/app/cosmic.rs
@@ -218,15 +218,14 @@ where
             self.app
                 .core()
                 .watch_config::<cosmic_theme::Theme>(
-                    if THEME
-                        .with(|t| {
-                            if let ThemeType::System { prefer_dark, .. } = t.borrow().theme_type {
-                                prefer_dark
-                            } else {
-                                None
-                            }
-                        })
-                        .unwrap_or_else(|| self.app.core().system_theme_mode.is_dark)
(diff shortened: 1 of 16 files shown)


mmstick:
There's a leftover thread local line at ` src/applet/mod.rs:436:10`

mmstick (approved):

wash2 (approved):

ids1024:
>  I am not sure it is possible to have access to this from multiple threads. 

The `SUBSURFACES` thread-local is a hack, since I couldn't find any other way for `Widget::draw` to pass back subsurface information to the iced-sctk shell. Ideally it should be some kind of local state / argument / return value.

If two threads are rendering separately they shouldn't be trying to share it, so yeah, it's correct to keep that one thread-local. Though I welcome any better ideas for handling it.

PR #573, opened 2024-08-13
Revert "fix: new matching algorithm for desktop entries"

Drakulix:
We should never use heuristics to match desktop-entries.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,6 @@ desktop = [
     "dep:freedesktop-desktop-entry",
     "dep:mime",
     "dep:shlex",
-    "dep:textdistance",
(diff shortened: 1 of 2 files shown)


wiiznokes:
I don't think this code was used anyway since it's now in https://github.com/pop-os/freedesktop-desktop-entry. I can see the security concern but we should at least keep a .contains imo. 

If you want to remove this feature from fde, it's just this function that need to be modified https://github.com/pop-os/freedesktop-desktop-entry/blob/b1967ab43c002fe1c23e52452ebb81985ab02637/src/matching.rs#L162

mmstick (approved):

PR #598, opened 2024-09-04
Minor spacing/padding fixes

git-f0x:
Removes the phantom vertical padding of the light and heavy horizontal dividers, caused by styling a `horizontal_rule` to a smaller width (e.g. 1.5 px of vertical padding on the light divider). This enables further design matching for apps and fixes off-center text in `list_column`. In the rare case that extra padding is required around a divider, a container can be used.

Matches the `spin_button` to the design.

Removes extra spacing from togglers, which caused them not be aligned with other widgets above/below in a list.

Reduces `list_column` spacing to match designs, but causes the height of some list items to be too low. E.g. the toggler has a height of 24, but in the designs, it's in a 32 height box, which makes list items without a description have less height than in the designs.
Not sure how to implement something like `min_height(32.0)` for `list_column` items.


mmstick:
> Not sure how to implement something like min_height(32.0) for list_column items.

When pushing elements into it, they could be wrapped with containers that set `height(32.0)`.

git-f0x:
But items that have a description seem to have a height of 34 (or more if the have multiple rows). Won't a container limit them to only 32.0?

mmstick:
Containers have both a `height` and `max_height` field, with `height` specifying the minimum.

git-f0x:
It still seems to be the same as the above screenshot with this: `self.children.push(container(item).height(32.0).max_height(256).into());`

git-f0x:
Is this something that's possible to do now, or are changes in Iced (or maybe the rebase) required for it to work?

mmstick:
We may need a custom list view widget implementation.

mmstick (approved):

PR #599, opened 2024-09-04
feat: allow access to ListColumn on Section

acolombier:
Add an helper method to edit Section's ListColumn properties on the fly

Used in https://github.com/pop-os/cosmic-settings/pull/557


mmstick:
I think it'd be better to have a function for passing an existing ListColumn into it

acolombier:
Updated as suggested!

src/widget/settings/section.rs
```diff
@@ -20,6 +20,25 @@ pub struct Section<'a, Message> {
...
+    /// 
+    /// ```
+    /// use cosmic::widget::{settings, text};
+    /// 
+    /// settings::view_section("My section")
+    ///   .list_column(ListColumn::default().spacing(42))
+    ///   .add(text("My item 1"))
+    ///   .add(text("My item 2"))
+    /// ```
+    pub fn list_column(mut self, list_column: ListColumn<'a, Message>) -> Self {
```
mmstick, line 36: I think this should be `with_list_column`

acolombier: Updated

mmstick (approved):
Thanks!

PR #601, opened 2024-09-07
feat: allow toggler to be disabled

acolombier:
Depends on #https://github.com/pop-os/iced/pull/169


src/widget/toggler.rs
```diff
@@ -7,7 +7,7 @@ use iced_core::text;
 pub fn toggler<'a, Message, Theme: iced_widget::toggler::StyleSheet, Renderer>(
     label: impl Into<Option<String>>,
     is_checked: bool,
-    f: impl Fn(bool) -> Message + 'a,
+    f: Option<impl Fn(bool) -> Message + 'a>,
```
mmstick, line 10: Since this is optional, it should be moved to a method

mmstick: It's fine to make breaking changes where necessary. There should be a method if we make the message optional. But we may want to wait for the iced rebase before we do this.

acolombier: I see so better hold off and go ahead with the current state? Or would you still want to see the `Toggler::on_change` method in that PR?

mmstick: I think we should wait for iced to be rebased before making new changes to iced

acolombier: Make sense. Would you suggest blocking this PR and dependencies then? Appreciate there currently seems to be [a blocker](https://github.com/pop-os/iced/pull/168) on iced rebase, so perhaps this change could go in as is, since it included minimal requirement for toggle matching an acceptable UX for async operation. 
Alternatively, thinking about the Bluetooth screen, I could put something like the `text("...")` with the toggler is not available. 

Appreciate all options are sub-optimal here, but just keen to get the ball rolling for COSMIC :)

PR #603, opened 2024-09-10
Interface density

git-f0x:
Since libcosmic seems to already have a Density enum, I moved most of the related things from Settings to here.

### DIFF
--- a/cosmic-theme/src/model/density.rs
+++ b/cosmic-theme/src/model/density.rs
@@ -1,0 +1,65 @@
+use crate::Spacing;
+use serde::{Deserialize, Serialize};
+
+#[derive(Copy, Clone, Debug, Default, Eq, PartialEq, Serialize, Deserialize)]
+pub enum Density {
+    Compact,
+    Spacious,
+    #[default]
+    Standard,
+}
+
+impl From<Density> for Spacing {
+    fn from(value: Density) -> Self {
+        match value {
+            Density::Compact => Spacing {
+                space_none: 0,
+                space_xxxs: 4,
+                space_xxs: 4,
+                space_xs: 8,
+                space_s: 8,
+                space_m: 16,
+                space_l: 24,
+                space_xl: 32,
+                space_xxl: 48,
+                space_xxxl: 64,
+            },
+            Density::Spacious => Spacing {
+                space_none: 4,
+                space_xxxs: 8,
+                space_xxs: 12,
+                space_xs: 16,
+                space_s: 24,
+                space_m: 32,
+                space_l: 48,
+                space_xl: 64,
+                space_xxl: 128,
+                space_xxxl: 160,
+            },
+            Density::Standard => Spacing {
+                space_none: 0,
+                space_xxxs: 4,
+                space_xxs: 8,
+                space_xs: 12,
+                space_s: 16,
+                space_m: 24,
+                space_l: 32,
+                space_xl: 48,
+                space_xxl: 64,
+                space_xxxl: 128,
+            },
+        }
+    }
+}
+
+impl From<Spacing> for Density {
+    fn from(value: Spacing) -> Self {
+        if value.space_m.saturating_sub(16) == 0 {
+            Self::Compact
+        } else if value.space_m.saturating_sub(24) == 0 {
+            Self::Standard
+        } else {
+            Self::Spacious
+        }
+    }
+}
--- a/src/widget/header_bar.rs
+++ b/src/widget/header_bar.rs
@@ -1,7 +1,7 @@
 // Copyright 2022 System76 <info@system76.com>
 // SPDX-License-Identifier: MPL-2.0
 
-use crate::config::Density;
+use crate::cosmic_theme::Density;
 use crate::{ext::CollectionWidget, widget, Element};
 use apply::Apply;
 use derive_setters::Setters;
@@ -278,8 +278,9 @@ impl<'a, Message: Clone + 'static> HeaderBar<'a, Message> {
         end.push(self.window_controls());
 
         let (height, padding) = match self.density.unwrap_or_else(crate::config::header_size) {
-            crate::config::Density::Compact => (36.0, 2.0),
-            crate::config::Density::Standard => (48.0, 8.0),
+            Density::Compact => (36.0, 2.0),
+            Density::Spacious => (48.0, 8.0),
+            Density::Standard => (48.0, 8.0),
         };
 
         // Creates the headerbar widget.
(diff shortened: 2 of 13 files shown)


git-f0x:
I'm marking this ready for review, since it doesn't cause any visual changes, and I'm not sure if I'll be able to rebase this if the Iced rebase happens in the meantime.
Not sure why the header bars don't respond to density changes, so it would be nice to fix that before this is merged, if possible (but I haven't yet been able to find the cause).

git-f0x:
The biggest visual issue (that Compact makes stand out more), is that [this](https://github.com/pop-os/libcosmic/blob/f12de010ec2bd76ee46ef2a33642cf22e55ae87e/src/widget/list/column.rs#L43) needs to have a `min_height` of 32 (potentially less for Compact) to match designs, but I'm not sure how to implement that (as mentioned [here](https://github.com/pop-os/libcosmic/pull/598#issuecomment-2329428222)).

git-f0x:
Changing the `interface_density` config file doesn't seem to do anything. Should ThemeBuilder be changed to take interface density into account (possibly instead of spacing, or to set the spacing to the value of interface density)?

wash2:
> Changing the `interface_density` config file doesn't seem to do anything. Should ThemeBuilder be changed to take interface density into account (possibly instead of spacing, or to set the spacing to the value of interface density)?

Ya the builder could have a method that sets spacing based on the provided density value. And the settings app could set the value.

git-f0x:
How would the builder access the `interface_density` config file (so that it works without Settings)? I tried importing CosmicTk to `/cosmic-theme/src/model/theme.rs` but I only managed to get circular dependencies or "not found".

git-f0x:
Settings already sets the spacing of the theme builder to the chosen density, but the config file isn't used for anything (though I made it write the density to the file).
So should I just move Density to cosmic-theme, and leave everything else as is?
Also regarding [this](https://github.com/pop-os/libcosmic/blob/914d989049ea1a8c47f0690cdbbfe3f862cc491d/src/app/mod.rs#L684), how would I access the current density, so that the padding and spacing could match the header bar (e.g. `padding = if Compact { 2 } else { 8 }`)?

git-f0x:
The NavBar side looks slightly bad with only 2 as padding, but not sure if the header bar horizontal padding should be increased (so that this can be increased as well).
screenshot-2024-09-18-18-05-48

Removing the conflicts caused by recent commits in Settings will take a bit of time.

cosmic-theme/src/model/density.rs
```diff
@@ -0,0 +1,65 @@
...
+                space_xxl: 64,
+                space_xxxl: 128,
+            },
+        }
+    }
+}
+
+impl From<Spacing> for Density {
+    fn from(value: Spacing) -> Self {
+        if (value.space_m - 16) < 1 {
```
wash2, line 57:
```suggestion
        if value.space_m.saturating_sub(16) == 0 {
```

cosmic-theme/src/model/density.rs
```diff
@@ -0,0 +1,65 @@
...
+            Self::Compact
+        } else if (value.space_m - 24) < 1 {
```
wash2, line 59:
```suggestion
        } else if value.space_m.saturating_sub(24) == 0 {
```

src/widget/header_bar.rs
```diff
@@ -361,7 +362,11 @@ impl<'a, Message: Clone + 'static> HeaderBar<'a, Message> {
         macro_rules! icon {
             ($name:expr, $size:expr, $on_press:expr) => {{
                 #[cfg(target_os = "linux")]
-                let icon = { widget::icon::from_name($name).apply(widget::button::icon) };
+                let icon = {
+                    widget::icon::from_name($name)
+                        .apply(widget::button::icon)
+                        .padding(8)
```
wash2, line 368: Does this padding not depend on density?

src/widget/header_bar.rs
```diff
@@ -372,6 +377,7 @@ impl<'a, Message: Clone + 'static> HeaderBar<'a, Message> {
                     )))
                     .symbolic(true)
                     .apply(widget::button::icon)
+                    .padding(8)
```
wash2, line 380: Does this padding not depend on density?

wash2 (changes requested):

src/widget/header_bar.rs
git-f0x: It does, but the default padding from the widget makes the buttons in the header too small when Compact, and cut off when Spacious.
They still perfectly fit inside the header when Compact, and I think match the buttons in server side decorations, which seem to have the same header height as Compact.

wash2:
> Should `interface_density` be removed from CosmicTk, since it isn't used anywhere, and the config file maybe moved to the builder (possibly also `header_size`)? Not sure what the `interface_density` config file can be used for, since it shouldn't be used by the builder to automatically set spacing so as to not override custom spacing values set by users, and the spacing can be set without it in Settings (`self.theme_builder.spacing = self.density.into()`).

I'm not sure what you mean. https://pop-os.github.io/libcosmic/cosmic/config/fn.interface_density.html returns the CosmicTk interface density value, and is used in your changes. And https://pop-os.github.io/libcosmic/cosmic/config/fn.header_size.html is used too.

git-f0x:
Oh, right, the config file is used in one place. Sorry for the needless question.
Those functions could maybe be moved to the builder (or elsewhere in cosmic-theme), since density is also in cosmic-theme. But probably not necessary.

git-f0x:
Sorry for bothering you, but could this potentially be merged soon, so that apps can update their libcosmic version for easier testing of what needs tweaks (and so that the Settings PR doesn't pull from my fork)?
The one thing I'm not entirely sure about is the way of getting the spacing values. I've seen something like this used in some places:
```
let guard = crate::theme::THEME.lock().unwrap();
let theme = guard.cosmic();
```
and in others something like `theme::active()::cosmic()`. Not sure what it optimal.

mmstick:
We do need some functions to make `cosmic::theme::active().cosmic().spacing` more ergonomic.

wash2 (dismissed):

git-f0x:
Oh, I didn't know that adding a commit automatically dismisses a review. This just fixed the icons in text input becoming cut off or way to small with density changes.

wash2 (approved):

wash2:
Could you fix the conflicts?

git-f0x:
It says on my end that it has no conflicts. So not sure what to do.

wash2 (approved):

PR #623, opened 2024-10-01
Compact density padding fixes

### LINKED ISSUE #619: Compact mode > No space to the left of the NavMenu

### LINKED ISSUE #620: Comapct mode > header bar items missing padding
Buttons (window controls and sidebar toggle) in headerbar are missing padding in compact mode.
screenshot-2024-09-27-08-04-07
Also search fields missing top padding
screenshot-2024-09-27-08-05-38
Compared to menubar highlight that has top padding
screenshot-2024-09-27-08-04-16

git-f0x:
Increases Compact header height to 40 (from 36), to better accommodate larger widgets (e.g. the search bar). This makes it distinct from the SSD header, which is now addressed by the `_` match case.
This needs a change in cosmic-comp at [this line](https://github.com/pop-os/cosmic-comp/blob/afdb65677857946b403043f04ca083810639e4e5/src/shell/element/window.rs#L506), in order to keep the current SSD style when libcosmic is updated there.

Also makes the horizontal header padding fixed at 8, so that the app window padding can also be 8.

Fixes #620.
Fixes #619.
(also https://github.com/pop-os/cosmic-epoch/issues/523 can probably be closed, since the Compact header exists).

Feel free to suggest improvements to the Density documentation wording (just added it to remove a warning).

### DIFF
--- a/src/widget/header_bar.rs
+++ b/src/widget/header_bar.rs
@@ -277,10 +277,10 @@ impl<'a, Message: Clone + 'static> HeaderBar<'a, Message> {
         end.push(widget::horizontal_space(Length::Fixed(12.0)).into());
         end.push(self.window_controls());
 
-        let (height, padding) = match self.density.unwrap_or_else(crate::config::header_size) {
-            Density::Compact => (36.0, 2.0),
-            Density::Spacious => (48.0, 8.0),
-            Density::Standard => (48.0, 8.0),
+        let height = match self.density.unwrap_or_else(crate::config::header_size) {
+            Density::Compact => 40.0,
+            Density::Spacious => 48.0,
+            Density::Standard => 48.0,
         };
 
         // Creates the headerbar widget.
@@ -316,8 +316,8 @@ impl<'a, Message: Clone + 'static> HeaderBar<'a, Message> {
             )
             .align_items(iced::Alignment::Center)
             .height(Length::Fixed(height))
-            .padding(padding)
-            .spacing(padding)
+            .padding([0, 8])
+            .spacing(8)
             .apply(widget::container)
             .style(crate::theme::Container::HeaderBar {
                 focused: self.focused,
@@ -387,13 +387,6 @@ impl<'a, Message: Clone + 'static> HeaderBar<'a, Message> {
             }};
         }
 
-        let density = self.density.unwrap_or_else(crate::config::header_size);
-        let spacing = if matches!(density, Density::Compact) {
-            2
-        } else {
-            8
-        };
-
         widget::row::with_capacity(3)
             .push_maybe(
                 self.on_minimize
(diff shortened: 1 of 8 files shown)


src/widget/header_bar.rs
```diff
@@ -278,9 +278,10 @@ impl<'a, Message: Clone + 'static> HeaderBar<'a, Message> {
         end.push(self.window_controls());
 
         let (height, padding) = match self.density.unwrap_or_else(crate::config::header_size) {
-            Density::Compact => (36.0, 2.0),
-            Density::Spacious => (48.0, 8.0),
-            Density::Standard => (48.0, 8.0),
+            Density::Compact => (40.0, 8),
+            Density::Spacious => (48.0, 8),
+            Density::Standard => (48.0, 8),
+            _ => (36.0, 2),
```
mmstick, line 284: Which variant would cosmic-comp use to get the `_` path?

git-f0x: Testing locally, just removing the `.density` line in comp makes it use the height of 36 (removing the line might make the density `None`, so it flows into the `_` case).

git-f0x: An alternative option to keep SSD and Compact CSD height the same would be to somehow reduce the height of the search bar when Compact. But I wasn't able to figure out if that's possible.

mmstick: We could treat None as server-side, or add a `server_side` bool, and I'm sure the layout method for text_input could be tweaked when compact.

git-f0x: Would it potentially be better to just keep the Compact header height as is, with no changes in cosmic-comp, and just change the icon padding of the text_input depending on density? I tested by reducing the padding of `system-search-symbolic` and `edit-clear-symbolic` to 4, and it can properly fit in the 36 height header.

git-f0x:
I think this might be good enough for now, until the designs for header bar densities are ready. It mostly reverts some of my changes in #603, and fixes the most obvious header and padding issues.

src/widget/header_bar.rs
git-f0x: I think it might be best to wait on differentiating between SSD and Compact CSD until the designs are ready (if they are designed as identical in height and padding, there wouldn't be any changes required).

git-f0x: Oh, actually, `cosmic-comp` seems to have `SSD_HEIGHT` hardcoded [here](https://github.com/pop-os/cosmic-comp/blob/a96394f7a6fd1d96289e319650e3f17b873cbfb0/src/shell/element/surface.rs#L96), so maybe the SSD-specific changes aren't necessary.

git-f0x:
I think this is ready as is, since SSD height is set in `cosmic-comp`, so no changes are required here.
Further changes will probably be made once Density designs are done (header heights, app padding, etc...).

mmstick (approved):

git-f0x:
Oh, I just noticed that the commit comment is out-of-date and referencing the `_` case. Sorry about that.

PR #628, opened 2024-10-04
Add font configuration capability

mmstick:
This also adds an optimization for the dropdown widget which significantly improves rendering performance when there's a lot of dropdown selections.

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -6,14 +6,16 @@
 use crate::cosmic_theme::Density;
 use cosmic_config::cosmic_config_derive::CosmicConfigEntry;
 use cosmic_config::{Config, CosmicConfigEntry};
+use iced::font::Family;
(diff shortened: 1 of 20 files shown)


src/config/mod.rs
```diff
@@ -94,6 +112,18 @@ impl Default for CosmicTk {
             icon_theme: String::from("Cosmic"),
             header_size: Density::Standard,
             interface_density: Density::Standard,
+            interface_font: FontConfig {
+                family: Ustr::from("Inter"),
```
jackpot51, line 116: At this time the defaults should be Fira and Fira Mono

jackpot51 (commented):
Can fira fonts still be bundled? One way to do that is to have the embedded fonts added to the global FontSystem

jackpot51 (approved):

PR #631, opened 2024-10-08
Fix color of horizontal segmented button divider

git-f0x:
This makes the `small_widget_divider` color match the designs, and thus fixes the divider color in the horizontal `segmented_button`. Also tweaks the values of `neutral_9` to match the designs (the diff is a bit hard to see).

The background color of the horizontal segmented button doesn't match designs, but I haven't been able to figure out where the color of the `small_widget` is set.
It can mostly match if doing something like the following, but it should use the `small_widget` color:
```
SegmentedButton::Control => {
                let cosmic = self.cosmic();
                let mut color = cosmic.palette.neutral_7;
                color.alpha = 0.1;
                let active = horizontal::selection_active(cosmic, &container.component);
                let rad_m = cosmic.corner_radii.radius_m;
                let rad_0 = cosmic.corner_radii.radius_0;
                Appearance {
                    background: Some(Background::Color(color.into())),
```

### DIFF
--- a/cosmic-theme/src/model/theme.rs
+++ b/cosmic-theme/src/model/theme.rs
@@ -186,7 +186,9 @@ impl Theme {
     #[allow(clippy::doc_markdown)]
     /// get @small_widget_divider
     pub fn small_widget_divider(&self) -> Srgba {
-        self.palette.neutral_9
+        let mut neutral_9 = self.palette.neutral_9;
+        neutral_9.alpha = 0.2;
+        neutral_9
     }
 
     // Containers


wash2:
The palette does need to be updated to match the designs, but it's not just the `neutral_9` color. I think it would be better to update the entire palette all at once rather than one variable at a time. The change to the alpha of the divider is fine though.

git-f0x:
Should I update the rest of the colors to match the [Base Colors](https://www.figma.com/design/SkAtS5qlVOGQC0n895AHzX/Design-System-(WIP)?node-id=12468-49526&node-type=section&t=yJaZ2xPVP5NmXhFF-0)?

wash2:
> Should I update the rest of the colors to match the [Base Colors](https://www.figma.com/design/SkAtS5qlVOGQC0n895AHzX/Design-System-(WIP)?node-id=12468-49526&node-type=section&t=yJaZ2xPVP5NmXhFF-0)? Though it'll be a lot of dividing by 255. 😅 Edit: And some colors here don't seem to exist in the designs (e.g. `gray_3`). Should this be updated to have only the Base Colors from the designs (alongside the colors used by the accent selection in Settings)?
> 
> Or should I just revert that change for now?

It would probably be easiest just to revert that change for now.

git-f0x:
Oh, it still lists it as a change. The files are the same as on master.

>If you'd like, you could try updating colors in https://github.com/pop-os/libcosmic/tree/iced-0.13 instead

I'll try later this week. Is there an efficient way to get the needed color values, without needing to convert from hex to decimal, and dividing by 255?

wash2:
> Oh, it still lists it as a change. The files are the same as on master.
> 
> > If you'd like, you could try updating colors in https://github.com/pop-os/libcosmic/tree/iced-0.13 instead
> 
> I'll try later this week. Is there an efficient way to get the needed color values, without needing to convert from hex to decimal, and dividing by 255? And would maybe formatting the file a bit be desirable (one color per line)?

You can build the palette in rust and serialize it to get the color values. You should be able to create colors like so
`let c: Srgba = "ffffffff".parse::<Srgba<u8>>().unwrap().into();`

I don't think it is necessary to format the file.

git-f0x:
>You can build the palette in rust and serialize it to get the color values. You should be able to create colors like so `let c: Srgba = "ffffffff".parse::<Srgba<u8>>().unwrap().into();`

I'm not sure if I know how to do that. Do I need to create a file with color names and the corresponding Hex values from the designs (formatted in some way), and then pass it to some program to serialize it? Though that would still require a lot of typing and checking.

Do I need to do anything else with this PR? The palette isn't changed, it's just reported like it is (should have maybe been a separate commit).

wash2:
> Do I need to do anything else with this PR? The palette isn't changed, it's just reported like it is (should have maybe been a separate commit).

Yes, please do:
```
git reset --soft HEAD~
git reset
git add cosmic-theme/src/model/theme.rs
git commit -m "fix(small_widget_divider): fix alpha value" -m "This makes this color match the designs, and thus fixes the divider color in the horizontal `segmented_button`."
git push -f
```

> Do I need to create a file with color names and the corresponding Hex values from the designs (formatted in some way), and then pass it to some program to serialize it? Though that would still require a lot of typing and checking.

Yes, you could create a temporary example, or another project that uses libcosmic as a path dependency. You would just need to create the palette from the hex colors and then serialize it all at once to a file. It saves you having to calculate the conversion for every color.

wash2 (approved):

PR #633, opened 2024-10-08
Update color palette

git-f0x:
This updates the color palette to the latest one from the designs.
I double (maybe triple) checked every value, so they should hopefully be correct. Also rearranged to colors to be in the same order as in the designs.

Should the changes in `gtk4_output.rs` maybe use `ext_COLOR` instead of `accent_COLOR`, like the other colors below (or should those colors use accent)?


git-f0x:
I'll close this, since these commits were added to the branch.

PR #669, opened 2024-10-30
Add transparency and blur support

edfloreshz:
This PR adds both transparency and blur support in supported platforms, currently tested in KDE Plasma.

image

cc @wash2


wash2:
I think the methods added to iced are sufficient for now. 

Blur and transparency will eventually just be part of the theme, or the toolkit config, and since they are not supported by COSMIC compositor yet, I don't think it should be added.

edfloreshz:
Sounds good!

PR #675, opened 2024-11-02
Add about page

edfloreshz:
This PR implements the About page for COSMIC applications.

- Added `About` struct to store the application's information.
- Added `about` method in `Application` trait to provide information about the application.
- Added `about_view` in `Application` trait with default implementation to render the About page.
- Added optional dependency `open` to handle `on_press` events for URLs in the about page. 

image


src/app/mod.rs
```diff
@@ -589,6 +600,102 @@ where
         panic!("no view for window {id:?}");
     }
 
+    #[cfg(feature = "desktop")]
+    /// Provides information about the application.
+    fn about(&self) -> Option<About> {
```
mmstick, line 605: This should return `Option<&About>` so we don't need to recreate it on each view update.

edfloreshz: Should be fixed, added an example as well.

mmstick (dismissed):

mmstick:
PR needs to be rebased, and then it should be good.

mmstick (approved):

mmstick (dismissed):

mmstick:
Needs to drop the change updating iced since that's causing a conflict

mmstick (approved):

PR #679, opened 2024-11-03
Add window border

git-f0x:
This adds a border around windows, and removes the 8px padding around the content, allowing the scrollbar to go to the edge (to match designs).
Also adds some invisible padding around non-maximized windows, so that the `resize_border` is more outside the window, to not interfere with the scrollbar.
This significantly improves the floating window resize UX, but currently also adds padding around tiled windows, so it should maybe be removed until there's a way to differently handle tiled windows (might already exist, but I haven't been able to find it), or if it's planned to do it in a better way than adding invisible padding (so that resize indicators don't show up when the cursor is at the edges of maximized windows, as they do now).
I tried to also add a window shadow, but it looked quite janky and artifacty, so that isn't here.

Moves the NavBar scroller to the edge of the NavBar, so that it doesn't overlap the buttons (at least when not Compact, so maybe the padding should be made fixed; or the scroller could be made similar to the one in `cosmic-edit` that reduces in width when not hovered over).

Setting the width of the scrollbar/scroller in the libcosmic `scrollable` seems to be hard to do properly, so maybe the default width should be updated in the Iced fork (diverging from upstream is a bit unfortunate, but it's a really small change).

This is `cosmic-files` with these changes (needs a few more changes in the app to match designs), and an updated scrollbar/scroller width in Iced:
screenshot-2024-11-03-23-19-25

### DIFF
--- a/examples/application/src/main.rs
+++ b/examples/application/src/main.rs
@@ -169,12 +169,12 @@ impl cosmic::Application for App {
             ]
             .width(iced::Length::Fill)
             .height(iced::Length::Shrink)
-            .align_x(iced::alignment::Horizontal::Center),
+            .align_x(iced::Alignment::Center),
         )
         .width(iced::Length::Fill)
         .height(iced::Length::Shrink)
-        .align_x(iced::alignment::Horizontal::Center)
-        .align_y(iced::alignment::Vertical::Center);
+        .align_x(iced::Alignment::Center)
+        .align_y(iced::Alignment::Center);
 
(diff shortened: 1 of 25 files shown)


mmstick (approved):

git-f0x:
I removed the invisible padding (since it caused the active hint to go around the padding) and added a (temporary) container around the main content, in order to move the scrollbar to its original place while keeping the other changes.

mmstick (approved):

PR #697, opened 2024-11-09
Move About to a widget

### LINKED ISSUE #696: Feedback on About feature
I find the api for about a bit odd. Imho it should be a widget. I don't understand why it is defined in the Application trait, since it's not something that libcosmic can position automatically. It would be more consistent to just call `widget::about(About {..})`. I don't really see an use case for overwridding `fn about()`.

@edfloreshz any thought ? i can work on it if its ok

edfloreshz:
Per @wiiznokes suggestion in #696, the About section code was moved to a widget.

Closes #696

### DIFF
--- a/src/widget/about.rs
+++ b/src/widget/about.rs
@@ -1,5 +1,3 @@
-use iced_widget::tooltip::Position;
-use license::License;
 #[cfg(feature = "desktop")]
 use {
     crate::{
@@ -7,7 +5,7 @@ use {
         widget::{self, horizontal_space},
         Element,
     },
-    std::collections::BTreeMap,
+    license::License,
 };
 
 #[derive(Debug, Default, Clone, derive_setters::Setters)]
@@ -30,28 +28,28 @@ pub struct About {
     license: Option<String>,
     /// Artists who contributed to the application.
     #[setters(skip)]
-    artists: BTreeMap<String, String>,
+    artists: Vec<(String, String)>,
     /// Designers who contributed to the application.
     #[setters(skip)]
-    designers: BTreeMap<String, String>,
+    designers: Vec<(String, String)>,
     /// Developers who contributed to the application.
     #[setters(skip)]
-    developers: BTreeMap<String, String>,
+    developers: Vec<(String, String)>,
     /// Documenters who contributed to the application.
     #[setters(skip)]
-    documenters: BTreeMap<String, String>,
+    documenters: Vec<(String, String)>,
     /// Translators who contributed to the application.
     #[setters(skip)]
-    translators: BTreeMap<String, String>,
+    translators: Vec<(String, String)>,
     /// Links associated with the application.
     #[setters(skip)]
-    links: BTreeMap<String, String>,
+    links: Vec<(String, String)>,
 }
 
 impl<'a> About {
     /// Artists who contributed to the application.
-    pub fn artists(mut self, artists: impl Into<BTreeMap<&'a str, &'a str>>) -> Self {
-        let artists: BTreeMap<&'a str, &'a str> = artists.into();
+    pub fn artists(mut self, artists: impl Into<Vec<(&'a str, &'a str)>>) -> Self {
+        let artists: Vec<(&'a str, &'a str)> = artists.into();
         self.artists = artists
             .into_iter()
             .map(|(k, v)| (k.to_string(), format!("mailto:{v}")))
@@ -60,8 +58,8 @@ impl<'a> About {
     }
 
     /// Designers who contributed to the application.
-    pub fn designers(mut self, designers: impl Into<BTreeMap<&'a str, &'a str>>) -> Self {
-        let designers: BTreeMap<&'a str, &'a str> = designers.into();
+    pub fn designers(mut self, designers: impl Into<Vec<(&'a str, &'a str)>>) -> Self {
+        let designers: Vec<(&'a str, &'a str)> = designers.into();
         self.designers = designers
             .into_iter()
             .map(|(k, v)| (k.to_string(), format!("mailto:{v}")))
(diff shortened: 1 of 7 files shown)


mmstick (approved):

mmstick:
Some tests are failing with an unresolved import for `about`

edfloreshz:
Sorry, should be fixed.

src/widget/mod.rs
```diff
@@ -365,3 +365,9 @@ pub use warning::*;
 #[cfg(feature = "markdown")]
 #[doc(inline)]
 pub use iced::widget::markdown;
+
+#[cfg(feature = "desktop")]
+pub mod about;
+#[cfg(feature = "desktop")]
+#[doc(inline)]
+pub use about::{about, About};
```
wiiznokes, line 373:
```suggestion
pub use about::about;
```

src/widget/about.rs
```diff
@@ -117,3 +125,81 @@ impl About {
         self
     }
 }
+
+#[cfg(feature = "desktop")]
```
wiiznokes, line 129: this can be removed

src/widget/about.rs
```diff
@@ -1,9 +1,17 @@
...
+use {
+    crate::{
+        iced::{alignment::Vertical, Alignment, Length},
+        widget::{self, horizontal_space},
+        Element,
+    },
+    std::collections::BTreeMap,
+};
 
 #[cfg(feature = "desktop")]
```
wiiznokes, line 11: this can be removed

src/widget/mod.rs
wiiznokes: nevermind this, it's not a widget

edfloreshz: I actually think this is a good idea.

src/widget/about.rs
edfloreshz: Removed!

src/widget/about.rs
edfloreshz: Removed!

wiiznokes:
some nitpicks: 
- would be nice if the fields were private, and the `set_` prefix were removed.
- some data are duplicated. For example, repository_url is stored in a dedicated field and the links BtreeMap. I can see the value of having dedicated fields, for example, to add custom icons in the future
- would be nice to add a method `add_link(self, name, link) -> Self`
- `set_support_url` is ambigus imo, for example, i have this value set in metainfo.xml: `bugtracker`, `donation`, `vcs-browser`, `homepage`
- maybe switch `license` with `license_type`, and rename `license_type` to `licence_text`

wdyt @edfloreshz  ?

edfloreshz:
Those are great suggestions, I'll make the changes in the next couple of hours.

wiiznokes:
here are some ideas of urls we can add https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines/#url

edfloreshz:
- Properties are now private.
- Removed the URL specific methods and made method `links` public, letting anyone pass an array of label/link pairs.
- Added the `license` crate to handle licenses. 
- Removed `license_type` and replaced it with a link to the license page (https://spdx.org/licenses).

@wiiznokes

src/widget/about.rs
```diff
@@ -0,0 +1,195 @@
...
+    icon: Option<String>,
+    /// The application’s version.
+    version: Option<String>,
+    /// Name of the application's author.
+    author: Option<String>,
+    /// Comments about the application.
+    comments: Option<String>,
+    /// The application's copyright.
+    copyright: Option<String>,
+    /// The license text.
```
wiiznokes, line 29:
```suggestion
    /// The license name.
```

src/widget/about.rs
```diff
@@ -0,0 +1,195 @@
...
+    license: Option<String>,
+    /// Artists who contributed to the application.
```
wiiznokes, line 31: Documentation for fields with `#[setters(skip)]` should be moved to the setter function instead.

src/widget/about.rs
```diff
@@ -0,0 +1,195 @@
...
+            .map(|(k, v)| (k.to_string(), format!("mailto:{v}")))
+            .collect();
+        self
+    }
+
+    pub fn developers(mut self, developers: impl Into<BTreeMap<&'a str, &'a str>>) -> Self {
+        let developers: BTreeMap<&'a str, &'a str> = developers.into();
+        self.developers = developers
+            .into_iter()
+            .map(|(k, v)| (k.to_string(), format!("mailto:{v}")))
```
wiiznokes, line 83: Could we provide a way to link something else than a email address ? Or just nothing. (maybe by checking if an `@` is present, or if the string is empty)

wiiznokes (approved):
I like the new api!
Just some minor doc issues

src/widget/about.rs
edfloreshz: Those fields are meant for email addresses, if the user needs to provide a URL they can with `links`.

wiiznokes:
Could you add this method to support the `fl!` macro ?

```rs
pub fn link(mut self, name: impl Into<String>, link: &'a str) -> Self {
    self.links.insert(name.into(), link.to_string());
    self
}
```

edfloreshz:
I'm thinking of the best way to support localization with the existing set of methods.

wiiznokes:
> I'm thinking of the best way to support localization with the existing set of methods.

smtg like that could work:

```rs
pub fn links<const N: usize>(
            mut self,
            links: [(impl Into<String>, impl Into<String>); N],
        ) -> Self {
            self.links = links
                .into_iter()
                .map(|(k, v)| (k.into(), v.into()))
                .collect();
            self
        }
```

edfloreshz:
Decided to do this instead:

```rust
pub fn links<T: Into<String>>(mut self, links: impl Into<Vec<(T, &'a str)>>) -> Self {
    let links: Vec<(T, &'a str)> = links.into();
    self.links = links
        .into_iter()
        .map(|(k, v)| (k.into(), v.to_string()))
        .collect();
    self
}
```

wiiznokes:
Yeah, it works. The `Vec` also make probably more sense.

I noticed that the scrollbar is a bit off, but idk if it's related or not. Maybe we need to adjust the padding.

Capture d’écran du 2024-11-09 21-08-58

edfloreshz:
Yeah we could give it some right padding

git-f0x:
Is there a reason that the scrollable was removed from the drawer? Since now it isn't possible to make the scrollbar be at the right edge (to match designs), and that spacing required to make the scrollbar not cover content makes things off-center.

wiiznokes:
> @wiiznokes Please confirm if I can ask Michael to review.

Yeah, go ahead

git-f0x:
If #698 gets merged, the scrollbar spacing can be removed, but `.padding([0, space_l, space_l, space_l])` should be applied to the column (before the scrollable).

mmstick (approved):

PR #698, opened 2024-11-09
[Context Drawer] Add actions to header

### LINKED ISSUE #625: libcosmic update for ContextDrawer header (back/forward/gallery)

git-f0x:
This adds configurable actions to the left side of the context drawer header. Not sure if it would be better to make the title optional, since it needs to be wrapped in `text::heading()`, so it would need a similar check regardless.
Maybe fixes #625.
screenshot-2024-11-09-19-37-20

Not quite sure why the scrollable for context drawer content was removed, since even those with only 1 item have to have a scrollbar at minimum vertical size. Every app will now have to also add `.padding([0, space_l, space_l, space_l])` to the content being added to a context drawer for each context drawer, in order for the scrollbar (if present) to be next to the edge of the drawer.
The second commit removes the built-in padding.

### DIFF
--- a/src/widget/context_drawer/widget.rs
+++ b/src/widget/context_drawer/widget.rs
@@ -67,7 +67,6 @@ impl<'a, Message: Clone + 'static> ContextDrawer<'a, Message> {
 
         let pane = column::with_capacity(2).push(header).push(
             container(drawer.into())
-                .padding([0, space_l, space_l, space_l])
                 .height(Length::Fill)
                 .width(Length::Shrink),
         );
--- a/src/widget/context_drawer/widget.rs
+++ b/src/widget/context_drawer/widget.rs
@@ -1,9 +1,7 @@
 // Copyright 2023 System76 <info@system76.com>
 // SPDX-License-Identifier: MPL-2.0
 
-use crate::widget::{
-    button, column, container, icon, row, scrollable, text, LayerContainer, Space,
-};
+use crate::widget::{button, column, container, icon, row, text, LayerContainer};
 use crate::{Apply, Element, Renderer, Theme};
 
 use super::overlay::Overlay;
@@ -26,7 +24,8 @@ pub struct ContextDrawer<'a, Message> {
 
 impl<'a, Message: Clone + 'static> ContextDrawer<'a, Message> {
     pub fn new_inner<Drawer>(
-        header: &'a str,
+        title: &'a str,
+        actions: Vec<Element<'a, Message>>,
         drawer: Drawer,
         on_close: Message,
         max_width: f32,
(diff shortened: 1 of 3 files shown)


src/widget/context_drawer/widget.rs
```diff
@@ -78,7 +89,8 @@ impl<'a, Message: Clone + 'static> ContextDrawer<'a, Message> {
 
     /// Creates an empty [`ContextDrawer`].
     pub fn new<Content, Drawer>(
-        header: &'a str,
+        title: &'a str,
+        actions: Vec<Element<'a, Message>>,
```
mmstick, line 93: Could `actions` be an optional setter?

git-f0x: Do you mean to just wrap the type in `Option<>`? It probably could, but it defaults to an empty Vec, so there's no need to manually specify anything in apps, if actions aren't desired.
Or are there plans to be able to put something else at the start of the header, and then choose one of multiple things to put at the start of the header?

mmstick: It's fine to create an empty Vec, with a method for pushing actions to it.

git-f0x:
```
fn context_header_actions(&self) -> Vec<Element<message::Message<Self::Message>>> {
        match &self.context_page {
            ContextPage::Preview(entity_opt, _kind) => {
                let entity = entity_opt.unwrap_or_else(|| self.tab_model.active());
                if let Some(tab) = self.tab_model.data::<Tab>(entity) {
                    if let Some(items) = tab.items_opt() {
                        for item in items.iter() {
                            if item.selected {
                                let mut actions = Vec::new();
                                if let Some(_path) = item.path_opt() {
                                    actions.extend(vec![
                                        widget::button::icon(widget::icon::from_name(
                                            "go-previous-symbolic",
                                        ))
                                        .on_press(message::app(Message::TabMessage(
                                            None,
                                            tab::Message::ItemLeft,
                                        )))
                                        .into(),
                                        widget::button::icon(widget::icon::from_name(
                                            "go-next-symbolic",
                                        ))
                                        .on_press(message::app(Message::TabMessage(
                                            None,
                                            tab::Message::ItemRight,
                                        )))
                                        .into(),
                                    ]);

                                    if item.can_gallery() {
                                        actions.push(
                                            widget::button::icon(widget::icon::from_name(
                                                "view-fullscreen-symbolic",
                                            ))
                                            .on_press(message::app(Message::TabMessage(
                                                None,
                                                tab::Message::Gallery(true),
                                            )))
                                            .into(),
                                        );
                                    }
                                }
                                return actions;
                            }
                        }
                    }
                }
            }
            _ => {}
        }
        Vec::new()
    }
```
This is the current code in Files to add actions to the header (a bit messy, could possibly be simplified). It just creates a Vec, and then that Vec is used in the header.
Would a method similar to `set_context_title` be what is wanted? E.g. `push_context_header_action(widget::button...)`.
Though it's a bit late here, so I won't be able to do it now.

mmstick: That would be fine

git-f0x: I'm a bit stumped as to how to add the method. Putting the actions Vec in Core doesn't really work, since it would require extensive changes, and if making a struct for it in `mod.rs`, I'm not sure how to have apps populate the Vec, while also reading that modified Vec in the `context_drawer` call in `mod.rs`.

mmstick (approved):

PR #699, opened 2024-11-10
[Context drawer] Add optional header and footer element

git-f0x:
This adds an optional element placed below the title row, so that the rest of the content can be wrapped in a scrollable.
Example of the current (likely suboptimal) API:
```
fn context_header_opt(&self) -> Option<Element<message::Message<Self::Message>>> {
        match self.context_page {
            ContextPage::NetworkDrive => Some(
                widget::column::with_capacity(2)
                    .spacing(space_m)
                    .push(widget::text_input(
                        fl!("enter-server-address"),
                        &self.network_drive_input,
                    ))
                    .push(widget::text(fl!("network-drive-description")))
                    .into(),
            ),
            _ => None,
        }
    }
```
Draft until a possibly better API is devised, if feasible.

### DIFF
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -71,13 +71,6 @@ use {
     zbus::{interface, proxy, zvariant::Value},
 };
 
-#[cfg(feature = "desktop")]
-use {
-    crate::widget,
-    iced::{alignment::Vertical, Alignment},
-    std::collections::BTreeMap,
-};
-
 pub(crate) fn iced_settings<App: Application>(
     settings: Settings,
     flags: App::Flags,
@@ -470,6 +463,16 @@ where
         Vec::new()
     }
 
+    /// Non-scrolling elements placed below the context drawer title row
+    fn context_drawer_header(&self) -> Option<Element<Message<Self::Message>>> {
+        None
+    }
+
+    /// Elements placed below the context drawer scrollable
+    fn context_drawer_footer(&self) -> Option<Element<Message<Self::Message>>> {
+        None
+    }
+
     /// Displays a dialog in the center of the application window when `Some`.
     fn dialog(&self) -> Option<Element<Self::Message>> {
         None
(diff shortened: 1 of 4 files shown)


mmstick:
A footer would also be required since we often have static buttons underneath scrollable lists

src/app/mod.rs
```diff
@@ -470,6 +463,10 @@ where
         Vec::new()
     }
 
+    fn context_header_opt(&self) -> Option<Element<Message<Self::Message>>> {
```
mmstick, line 466: I think `context_drawer_header` would be better

wiiznokes:
Just throwing idea here:

Making a specific struct for this:

```rs
pub struct ContextDrawer<Message> {
    on_close: Option<fn() -> Message>,
    title: Option<String>,
    header_actions: Vec<Element<Message>>,
    // whatever we need
}
```

Adding just this function to the trait:

```rs
fn context_drawer(&self) -> Option<ContextDrawer<Self::Message>> {
    None
}
```

This way, the app code could look like this:

```rs
enum Drawer {
    About,
    Other
}

struct AppState {
    drawer: Option<Drawer>,
}

fn context_drawer(&self) -> Option<ContextDrawer<Self::Message>> {
    self.drawer.as_ref().map(|drawer| {
        match drawer {
            Drawer::About => {
                ContextDrawer::new()
                    .title("about")
                    .content(widget::about())
                    .on_close(Message::CloseDrawer)
            },
            Drawer::Other => todo!(),
        }
    })
}
```

Advantages:

- no need to remember to set `self.core.window.show_context`
- same goes for the drawer title i think
- debloat the `Application` trait, and make it easier to find the drawer related methods

git-f0x:
That seems like a nice improvement @wiiznokes! Though I'm not sure if I know how to implement it properly.
If someone wants to make those improvements, should I make this ready for review?

mmstick (approved):

wiiznokes:
I could try to implement this if this is wanted. @mmstick wdyt about this design ?

mmstick:
@wiiznokes Sure. I think that would help with the current complexity

PR #706, opened 2024-11-11
feat: ContextDrawer struct in Application trait

wiiznokes:
This pr implement the design from this comment https://github.com/pop-os/libcosmic/pull/699#issuecomment-2466828302.

It seems to works as expected, but i think it will cause problem, because in this design, the field `core.windows.show_context` is not meant to be changed. It is used in some calculation for responsiveness if i understand correctly.

Notably:
```rs
/// Call this whenever the scaling factor or window width has changed.
    #[allow(clippy::cast_precision_loss)]
    fn is_condensed_update(&mut self) {
        // Nav bar (280px) + padding (8px) + content (360px)
        let mut breakpoint = 280.0 + 8.0 + 360.0;
        //TODO: the app may return None from the context_drawer function even if show_context is true
        if self.window.show_context && !self.window.context_is_overlay {
            // Context drawer min width (344px) + padding (8px)
            breakpoint += 344.0 + 8.0;
        };
        self.is_condensed = (breakpoint * self.scale_factor) > self.window.width as f32;
        self.nav_bar_update();
    }
```

```rs
fn condensed_conflict(&self) -> bool {
        // There is a conflict if the view is condensed and both the nav bar and context drawer are open on the same layer
        self.is_condensed
            && self.nav_bar.toggled_condensed
            && self.window.show_context
            && !self.window.context_is_overlay
    }
```
and others.

I think it should be possible to only rely on the return type of `fn context_drawer(&self)` and `fn nav_bar(&self)` (and potentially others), to know how thing should be shown on screen, by moving this calculation in the view function.

cc @wash2 (i think some of this code come from you)

### DIFF
--- a/examples/about/src/main.rs
+++ b/examples/about/src/main.rs
@@ -3,7 +3,7 @@
 
 //! Application API example
 
-use cosmic::app::context_drawer::{self, context_drawer, ContextDrawer};
+use cosmic::app::context_drawer::{self, ContextDrawer};
 use cosmic::app::{Core, Settings, Task};
 use cosmic::iced::widget::column;
 use cosmic::iced_core::Size;
--- a/examples/about/src/main.rs
+++ b/examples/about/src/main.rs
@@ -110,8 +110,7 @@ impl cosmic::Application for App {
     fn update(&mut self, message: Self::Message) -> Task<Self::Message> {
         match message {
             Message::ToggleAbout => {
-                self.core.window.show_context = !self.core.window.show_context;
-                self.core.set_show_context(self.core.window.show_context);
+                self.set_show_context(!self.core.window.show_context);
                 self.show_about = !self.show_about;
             }
             Message::Open(url) => match open::that_detached(url) {
--- a/examples/about/src/main.rs
+++ b/examples/about/src/main.rs
@@ -3,6 +3,7 @@
 
 //! Application API example
 
+use cosmic::app::context_drawer::{context_drawer, ContextDrawer};
 use cosmic::app::{Core, Settings, Task};
 use cosmic::iced::widget::column;
 use cosmic::iced_core::Size;
@@ -35,6 +36,7 @@ pub struct App {
     core: Core,
     nav_model: nav_bar::Model,
     about: About,
+    show_about: bool,
 }
 
 /// Implement [`cosmic::Application`] to integrate with COSMIC.
@@ -80,6 +82,7 @@ impl cosmic::Application for App {
             core,
             nav_model,
             about,
+            show_about: false,
         };
 
         let command = app.update_title();
@@ -98,20 +101,18 @@ impl cosmic::Application for App {
         self.update_title()
     }
 
-    fn context_drawer(&self) -> Option<Element<Self::Message>> {
-        if !self.core.window.show_context {
-            return None;
+    fn context_drawer(&self) -> Option<ContextDrawer<Self::Message>> {
+        match self.show_about {
+            true => Some(context_drawer(widget::about(&self.about, Message::Open), Message::ToggleAbout)),
+            false => None,
         }
-
-        Some(widget::about(&self.about, Message::Open))
     }
 
     /// Handle application events here.
     fn update(&mut self, message: Self::Message) -> Task<Self::Message> {
         match message {
             Message::ToggleAbout => {
-                self.core.window.show_context = !self.core.window.show_context;
-                self.core.set_show_context(self.core.window.show_context)
+                self.show_about = !self.show_about;
             }
             Message::Open(url) => match open::that_detached(url) {
                 Ok(_) => (),
--- a/examples/about/src/main.rs
+++ b/examples/about/src/main.rs
@@ -3,7 +3,7 @@
 
 //! Application API example
 
-use cosmic::app::context_drawer::{context_drawer, ContextDrawer};
+use cosmic::app::context_drawer::{self, context_drawer, ContextDrawer};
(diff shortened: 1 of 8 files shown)


examples/about/src/main.rs
```diff
@@ -98,20 +101,18 @@ impl cosmic::Application for App {
         self.update_title()
     }
 
-    fn context_drawer(&self) -> Option<Element<Self::Message>> {
-        if !self.core.window.show_context {
-            return None;
+    fn context_drawer(&self) -> Option<ContextDrawer<Self::Message>> {
+        match self.show_about {
+            true => Some(context_drawer(widget::about(&self.about, Message::Open), Message::ToggleAbout)),
```
mmstick, line 106: We could add a convenience function for this. Like so

```rs
Some(context_drawer::about(&self.about, Message::Open, Message::ToggleContext))
```

examples/about/src/main.rs
```diff
@@ -98,20 +101,18 @@ impl cosmic::Application for App {
...
-
-        Some(widget::about(&self.about, Message::Open))
     }
 
     /// Handle application events here.
     fn update(&mut self, message: Self::Message) -> Task<Self::Message> {
         match message {
             Message::ToggleAbout => {
-                self.core.window.show_context = !self.core.window.show_context;
-                self.core.set_show_context(self.core.window.show_context)
```
mmstick, line 114: I think this should still be defined. Maybe this could simplified with a convenience method on the application trait

```rs
self.toggle_show_context()
```

src/app/mod.rs
```diff
@@ -745,58 +725,69 @@ impl<App: Application> ApplicationExt for App {
             };
 
             if self.nav_model().is_none() || core.show_content() {
-                let main_content = self.view().map(Message::App);
+                let main_content = self.view();
 
                 //TODO: reduce duplication
                 let context_width = core.context_width(has_nav);
                 if core.window.context_is_overlay {
```
mmstick, line 732: This should also check if the context drawer is currently set to show.

examples/about/src/main.rs
```diff
@@ -98,20 +101,18 @@ impl cosmic::Application for App {
         self.update_title()
     }
 
-    fn context_drawer(&self) -> Option<Element<Self::Message>> {
-        if !self.core.window.show_context {
-            return None;
+    fn context_drawer(&self) -> Option<ContextDrawer<Self::Message>> {
+        match self.show_about {
```
mmstick, line 105: Better to use `self.show_about.then(|| {})`

examples/about/src/main.rs
wiiznokes: Why though? I think it is simpler to just rely on `context_drawer()` being `Some` or `None` to know if a context should be displayed. It could have many update branch were a context should be trigger. It also match the current documentation: 

> /// Displays a context drawer on the side of the application window when `Some`.

This would also be consistent with `dialog`, `footer` and other

wiiznokes: I will try to make a poc of this eventually. I think i see where this could cause problem now

mmstick: There are many internal behavioral side effects that libcosmic manages when the value is toggled. The context drawer affects the nav bar in certain situations, and vice versa nav bar updates affects the context drawer. Some of these are handled internally by libcosmic's application runtime.

src/app/mod.rs
wiiznokes: Can i ask what is `context_is_overlay` ?

mmstick: We recently made it possible to configure the context drawer to be part of the application view's layout for cosmic-files, rather than having it placed in a widget overlay.

examples/about/src/main.rs
wiiznokes: > I think this should still be defined. Maybe this could simplified with a convenience method on the application trait
> 
> ```rust
> self.toggle_show_context()
> ```

This one already existed. I added it to the doc.

mmstick:
I'll give it a try with cosmic-settings

src/widget/mod.rs
```diff
@@ -119,9 +119,9 @@ pub use color_picker::{ColorPicker, ColorPickerModel};
 #[doc(inline)]
 pub use iced::widget::qr_code;
 
-pub mod context_drawer;
+pub(crate) mod context_drawer;
 #[doc(inline)]
-pub use context_drawer::{context_drawer, ContextDrawer};
```
mmstick, line 124: These should remain public for applications that override the default view implementation.

src/app/mod.rs
```diff
@@ -745,58 +726,69 @@ impl<App: Application> ApplicationExt for App {
...
-                            context_drawer(
-                                &core.window.context_title,
-                                self.context_header_actions(),
-                                self.context_drawer_header(),
-                                self.context_drawer_footer(),
-                                Message::Cosmic(cosmic::Message::ContextDrawer(false)),
-                                main_content,
-                                context.map(Message::App),
-                                context_width,
+                            Element::from(
```
mmstick, line 736: I'd prefer to reduce the number of indentations where possible. Using either `.into()` or `.apply(Element::from)`

wiiznokes: Yeah it was due to the .map function. I agree apply is better

src/widget/mod.rs
wiiznokes: Ig in the future, we could directly use this widget instead of the custom struct, but i was unsure if some properties should be exposed

mmstick (approved):

PR #714, opened 2024-11-14
Vertical Spin Button Widget

bhh32:
# Vertical Spin Button Widget
Is like the Spin Box Button widget, but vertical instead of horizontal.

## Reason
I needed this for an application I was creating and would like to contribute as an additional widget for other developers to use.

## Screenshot
screenshot-2024-11-14-22-29-40


mmstick (changes requested):
Make sure that the API matches the existing spin_button widget: https://github.com/pop-os/libcosmic/blob/master/src/widget/spin_button/mod.rs

The widget's module should also be in the spin_button module.

bhh32:
> Make sure that the API matches the existing spin_button widget: https://github.com/pop-os/libcosmic/blob/master/src/widget/spin_button/mod.rs
> 
> The widget's module should also be in the spin_button module.

I've moved the vertical spin button's code into the spin_button's mod.rs. I also changed the struct name from VerticalSpinner to VerticalSpinButton to match that of the SpinButton widget.

bhh32 (commented):
Updated the VerticalSpinButton struct from VerticalSpinner and moved the code to spin_button->mod.rs to comply with the request to do so.

mmstick:
The API is still different. Make sure it matches the sign button widget. The parameters should be the same. It would be unexpected for them to have different behaviors.

bhh32:
I’m a little confused. I wanted to give the user of the widget the ability
to choose their own min and max, allowing it to be flexible. If I remove
those parameters from the new and vertical_spin_button functions users of
the widget lose that ability. I have not actually used the spin_button
widget because I needed a vertical one for what I was working on. Does it
not have the ability for the widget user to dictate the min and max value
of the widget? Based on the current API, I don’t see where it would be
possible since that functionality seems to come from the Model struct.

wiiznokes:
Personally, i found the slider implementation a bit cleaner https://github.com/iced-rs/iced/blob/2a2e20b0a35bcf4da9ac34ab2f664485c4e1dbe3/widget/src/slider.rs#L84.

Also, i think it would be better to use .on_press_maybe to have a disabled state when an incrementation is not possible.

bhh32:
> Personally, i found the slider implementation a bit cleaner https://github.com/iced-rs/iced/blob/2a2e20b0a35bcf4da9ac34ab2f664485c4e1dbe3/widget/src/slider.rs#L84.
> 
> 
> 
> Also, i think it would be better to use .on_press_maybe to have a disabled state when an incrementation is not possible.
> 
> 

Yes, I can get onboard with changing it up to have an inclusive range because it keeps the flexibility I was trying to provide. Same as the on_press_maybe and having the ability to be disabled, giving it even more flexibility.

This doesn't address the original request to make it like the current spin button API though. Would it be okay for me to tweak the current spin button to match the agreed upon flexibility of the vertical spin button so that they match? I ask because I agree the API for these two widgets should match for user experience.

I would basically just take away the model that current the spin button is using and add the functionality into the SpinButton struct and its functions like I did with VerticalSpinButton. They could share the increment and decrement functions I created. I'd just make a trait they both implement and the function generic with the restriction of that trait so they both can use the function.

I even had an idea that they could be the same widget, but I add a Direction enum that would be used in the signature that the developer could specify which type they want and the Element returned in the from function would be based on a match case coming from the this variable.

mmstick:
The existing spin button model already supports setting the value, min, max, and step.

bhh32:
> The existing spin button model already supports setting the value, min, max, and step.

Can you explain where in the API this can happen? I just am not seeing it, and I apologize for missing it. I'm having a hard time understanding how to match what I've done to the existing spin button's API. I see that the model has setters, but in the current API I don't see how and where they're being used.

mmstick:
See the Model struct. Each field has a setter. https://github.com/pop-os/libcosmic/blob/a355a049d917823dafcd6669cad6afa6983bc6c6/src/widget/spin_button/model.rs#L16-L30

Example: https://github.com/pop-os/libcosmic/blob/a355a049d917823dafcd6669cad6afa6983bc6c6/examples/cosmic/src/window/demo.rs#L128-L132

View: https://github.com/pop-os/libcosmic/blob/a355a049d917823dafcd6669cad6afa6983bc6c6/examples/cosmic/src/window/demo.rs#L270

Updating: https://github.com/pop-os/libcosmic/blob/a355a049d917823dafcd6669cad6afa6983bc6c6/examples/cosmic/src/window/demo.rs#L181-L186

Supports Decimal type

https://github.com/pop-os/libcosmic/blob/a355a049d917823dafcd6669cad6afa6983bc6c6/examples/cosmic/src/window/demo.rs#L108

bhh32:
> See the Model struct. Each field has a setter. https://github.com/pop-os/libcosmic/blob/a355a049d917823dafcd6669cad6afa6983bc6c6/src/widget/spin_button/model.rs#L16-L30
> 
> 
> 
> Example: https://github.com/pop-os/libcosmic/blob/a355a049d917823dafcd6669cad6afa6983bc6c6/examples/cosmic/src/window/demo.rs#L128-L132
> 
> 
> 
> View: https://github.com/pop-os/libcosmic/blob/a355a049d917823dafcd6669cad6afa6983bc6c6/examples/cosmic/src/window/demo.rs#L270
> 
> 
> 
> Updating: https://github.com/pop-os/libcosmic/blob/a355a049d917823dafcd6669cad6afa6983bc6c6/examples/cosmic/src/window/demo.rs#L181-L186
> 
> 
> 
> Supports Decimal type
> 
> 
> 
> https://github.com/pop-os/libcosmic/blob/a355a049d917823dafcd6669cad6afa6983bc6c6/examples/cosmic/src/window/demo.rs#L108
> 
> 
> 
> 

Thank you! That is definitely helpful in understanding how the current spin button works. However, I'm not understanding the logic of why it works that way.

From what I'm getting from that example is that you have the spin button widget that truly doesn't have its own functionality. It relies upon the Model struct to actually do anything. I'm not sure why it's built that way. The spin button doesn't have a way to intuitively set the min, max, and step. You setting the min, max, and step in the model, which you're also updating any values from.

My vertical spin button has it all included and there's no need to call a different struct to set any values. I personally, as we can see through this conversation, can't see how that's an easy to use and intuitive way of making an API for a widget. Especially when other widgets don't rely on a model struct to do anything, you just create it in the view function, send the message that you need to send to the update function to do any app state updates. The way it currently is makes it where the developer has to keep track of more things. 

If you have a look at my [example](https://github.com/bhh32/libcosmic/tree/vertical-spinner/examples/vertical-spin-button) for the vertical spin button, you can see how it is simplified from the way that the current spin button's functionality works.

PR #715, opened 2024-11-15
Fix footer corner radius

git-f0x:
This makes the primary container style use `radius_s`, instead of `radius_xs`, which matches the radius of the footer to the designs.
Also fixes the vertical padding of the `context_drawer` footer.

Since other container styles also used `radius_xs`, I also changed them for consistency.
Testing with `cosmic-files`, I didn't notice anything unusual from this change, but not sure if this should be done, or if the footer should use a custom style (though there isn't a `Custom` Layer enum).


wash2 (approved):
This seems to be the correct border radius for the context drawer too.

PR #718, opened 2024-11-17
Support launching app in terminal

jpttrssn:
- Get terminal setting from desktop file
- Launch app in terminal if set

This supports https://github.com/pop-os/cosmic-epoch/issues/735

Requires: https://github.com/pop-os/cosmic-term/pull/380


wash2 (changes requested):

src/desktop.rs
```diff
@@ -243,14 +249,23 @@ where
         _ => return,
     };
 
-    let mut cmd = std::process::Command::new(&executable);
-
-    for arg in exec {
-        // TODO handle "%" args here if necessary?
-        if !arg.starts_with('%') {
-            cmd.arg(arg);
+    let mut cmd = match terminal {
```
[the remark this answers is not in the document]
jpttrssn: Ah, yes, much better. Thank you!

src/desktop.rs
```diff
@@ -243,14 +249,20 @@ where
...
     };
 
-    let mut cmd = std::process::Command::new(&executable);
-
-    for arg in exec {
-        // TODO handle "%" args here if necessary?
-        if !arg.starts_with('%') {
-            cmd.arg(arg);
+    let mut cmd = if terminal {
+        let mut cmd = std::process::Command::new("cosmic-term");
```
jackpot51, line 253: What happens if cosmic-term is not installed?

jpttrssn: Good point. I'll try to come up with something. I see that `cosmic-files` handles this. And I assume the code for `command + t` also handles this but I haven't found that yet.

Cargo.toml
```diff
@@ -117,11 +121,13 @@ tokio = { version = "1.24.2", optional = true }
 tracing = "0.1.41"
 unicode-segmentation = "1.6"
 url = "2.4.0"
+xdg = { version = "2.5.2", optional = true }
 zbus = { version = "4.2.1", default-features = false, optional = true }
 
 [target.'cfg(unix)'.dependencies]
 freedesktop-icons = { package = "cosmic-freedesktop-icons", git = "https://github.com/pop-os/freedesktop-icons" }
 freedesktop-desktop-entry = { version = "0.5.1", optional = true }
+freedesktop_entry_parser = "1.3"
```
mmstick, line 130: Parsing of desktop entries is already provided by `freedesktop-desktop-entry` in the line above.

PR #720, opened 2024-11-17
Simplify and Enhance Spin Button Widget

bhh32:
# Purpose
The purpose of this PR is to simplify the Spin Button widget's API for future external COSMIC developers by removing the external Model struct that the previous version had. The Model struct made it to where the developer had to keep track of the functionality, initialization, and value updating outside of the Spin Button widget itself.  
  
It also is to expand on functionality as an Orientation enum was added to the API so that developers can choose between a Horizontal (default) or Vertical orientation.

## Spin Button Widget Improvements
The improvement over the previous implementation, although a breaking change, has all of the functionality of the Model struct self-contained within the Spin Button widget itself.  
  
This will allow for developers who are using the library to no longer have to keep track of the creation of the Spin Button widget and the functionality through an external Model struct.  
  
The messages passed by the Spin Button widget to the applications' update function, in the updated version, will allow developers to update the current value of the Spin Button widget through the application state itself without calling an external spin_button::model::update(value) function. See the below screenshots as an example of this:  
  
spin_button_vertical_view
  
spin_button_update_example  
  
As can be seen, the Spin Button widget is created in the application view function, and like other widgets, the value is updated in the application update function. This keeps the application state in tact and keeps with the design of most of the other widgets in the library.  
  
It can also be seen that the API signature is quite a bit more verbose. Although this can be cumbersome sometimes, I believe that it gives the developer more flexibility when the Spin Button widget is created, but doesn't have to be touched again after the creation itself and gives a one stop shop if something does need to be updated.  
  
Also, I've added the flexibility of an orientation field that comes from a new Orientation enum that has two variants, Horizontal and Vertical. This field is designed as an Option<Orientation> type so that there could be a default behavior and a developer doesn't need to import the spin_button::Orientation enum if they just want a default Spin Button widget. This can be achieved by putting None into the orientation field on creation. This can be seen in the following screenshot from the example:  
  
spin_button_horizontal_view  
  
If a developer needs a standard spin button, I've made the API little shorter adding a spin_button_standard function that does not take the orientation parameter and just passes None to the SpinButton::new() function as the orientation parameter.  
  
spin_button_standard_implementation
  
spin_button_standard_view
  
Another benefit in doing it this way is that the application state and messages can be created in a much cleaner manner, and only the initialization of the application state fields needs to happen. There is no Model struct initialization that needs to occur.  
  
spin_button_app_state_and_msgs  
  
The widget does inherit the theme of the application, just like all other widgets do. Here is a screenshot of the example created with this PR and what the new Spin Button looks like in both orientations.  
  
spin_button_example

Here is a link to the working [example](https://github.com/bhh32/libcosmic/tree/combined-spin-button/examples/spin-button) of the proposed Spin Button widget in this PR.  
  
Here is a link to the modified spin_button->[mod.rs](https://github.com/bhh32/libcosmic/blob/combined-spin-button/src/widget/spin_button/mod.rs) for review. It is heavily commented to mark out my thoughts and design choices as I was creating it.
  
As I realize that this is a breaking change, I am willing to help go through and resolve any issues that occur in the COSMIC default apps that would result if this change is accepted.
  
I’ve already identified a couple of places in cosmic-settings that would be affected by this and have forked cosmic-epoch to start working out how to repair what would be broken without affecting functionality.


examples/spin-button/src/main.rs
```diff
@@ -0,0 +1,193 @@
...
+
+    fn view(&self) -> Element<Self::Message> {
+        let vert_spinner_row = row![
+            spin_button(
+                "i8", // label: displayed above the widget no matter the orientation
+                1, // step: how much to increment/decrement by
+                self.i8_num, // current value, this is also what's displayed in the center of the widget
+                -5, // minimum value, if decremented below this the widget's current value rolls to the max value
+                5, // maximum value, if incremented above this the widget's current value rolls to the min value
+                Some(Orientation::Vertical), // oreintation of the widget
```
mmstick, line 94: The default spin_button widget should be the horizontal spin button. Instead of an enum, the vertical spin button should be its own function: `widget::spin_button::vertical()` / `widget::vertical_spin_button()`.

src/app/mod.rs
```diff
@@ -102,7 +102,7 @@ pub(crate) fn iced_settings<App: Application>(
     iced.default_font = settings.default_font;
     iced.default_text_size = iced::Pixels(settings.default_text_size);
     let exit_on_close = settings.exit_on_close;
-    iced.is_daemon = false;
```
mmstick, line 105: This shouldn't be touched

src/widget/mod.rs
```diff
@@ -311,7 +311,7 @@ pub mod settings;
 
 pub mod spin_button;
 #[doc(inline)]
-pub use spin_button::{spin_button, SpinButton};
+pub use spin_button::{spin_button, spin_button_standard, SpinButton};
```
mmstick, line 314: This should still be `spin_button`

iced
mmstick, line 1: PR shouldn't need to modify iced submodule

examples/spin-button/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "vertical-spin-button"
+version = "0.1.0"
+edition = "2021"
+
+[features]
+default = ["xdg-portal"]
+rfd = ["libcosmic/rfd"]
```
mmstick, line 8: Not used by example

src/app/mod.rs
bhh32: Yes, I agree. I only commented it out because the example wouldn't build due to the compiler saying it no longer existed.

src/widget/mod.rs
bhh32: I agree, I'll make this change.

src/widget/spin_button/mod.rs
```diff
@@ -1,108 +1,260 @@
 // Copyright 2022 System76 <info@system76.com>
 // SPDX-License-Identifier: MPL-2.0
 
-//! A control for incremental adjustments of a value.
-
-mod model;
-use std::borrow::Cow;
+// Updated by Bryan Hyland <bryan.hyland32@gmail.com>
+// Updated on: 17Nov24
```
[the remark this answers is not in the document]
bhh32: Yes, I forgot to pull them out.

examples/spin-button/Cargo.toml
bhh32: I will remove them.

iced
bhh32: I didn't actually modify anything, I just pulled it in using git submodule update --init --recursive so that it actually existed for me to use in my fork. Otherwise, it shows the directory to be empty. Is there another way I should have done this?

src/widget/spin_button/mod.rs
```diff
@@ -1,108 +1,259 @@
...
+        row, 
+        text,
+    },
+};
 
-pub struct SpinButton<'a, Message> {
-    label: Cow<'a, str>,
-    on_change: Box<dyn Fn(model::Message) -> Message + 'static>,
+#[derive(Clone, Copy)]
+pub enum Orientation {
```
mmstick, line 28: Since this is a private type, it shouldn't be `pub`

src/widget/mod.rs
```diff
@@ -311,7 +311,7 @@ pub mod settings;
 
 pub mod spin_button;
 #[doc(inline)]
-pub use spin_button::{spin_button, SpinButton};
+pub use spin_button::{spin_button, vertical_spin_button, SpinButton};
```
mmstick, line 314: `vertical as vertical_spin_button`

src/widget/spin_button/mod.rs
bhh32: I can't actually make the SpinButton struct private, which means I probably shouldn't make Orientation private either.

This is what happens when I try to make SpinButton private:  
  
spin_button_private_type

src/widget/spin_button/mod.rs
```diff
@@ -1,108 +1,259 @@
...
+    on_press: impl Fn(T) -> M + 'static,
+) -> SpinButton<'a, T, M>
+where 
+    T: Add<Output = T> + Sub<Output = T> + PartialEq + PartialOrd + Display + Copy
+{
+    SpinButton::new(label, step, value, min, max, None, on_press)
+}
+
+/// Shorthand to create a standard (horizontal) spin button widget
+pub fn vertical_spin_button<'a, T, M>(
```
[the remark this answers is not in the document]
bhh32: Updated to vertical from vertical_spin_button.

src/widget/mod.rs
bhh32: Updated to vertical in place of vertical_spin_button.

src/widget/spin_button/mod.rs
mmstick: The SpinButton struct should be public, but the Orientation enum does not need to be.

bhh32: Ok, I see where you're going. I made the new() function for SpinButton private as well that way I can change Orientation to enum private.

mmstick (dismissed):

mmstick:
Make sure to configure your editor to automatically run `cargo fmt` on save.

bhh32:
Fixed formatting issues that failed checks.

bhh32:
> Make sure to configure your editor to automatically run `cargo fmt` on save.

I just set that up, thank you. It's been fixed and pushed up.

mmstick:
`cargo fmt` still failing: https://github.com/pop-os/libcosmic/actions/runs/11900369391/job/33161645901?pr=720

bhh32:
Interesting, I made a small change, saved, and changed it back to have it run cargo fmt again. I've committed and pushed. I hope that that works.

mmstick (approved):

mmstick:
Thanks!

mmstick:
Note: had to fix widgets being named `vertical`

mmstick:
Removed the rollover behavior. It should be clippped, or otherwise implementation-defined by the user of the widget.

bhh32:
Thank you for accepting this! I really appreciate it. I'll put it on my list to add a setter or something to add the rollover feature since it's something I personally needed when I first needed to use a vertical spin button. Not a high priority at the moment though.

mmstick:
Need to revert to fix other issues

src/widget/spin_button/mod.rs
```diff
@@ -1,108 +1,276 @@
...
-    on_change: impl Fn(model::Message) -> Message + 'static,
-) -> SpinButton<'a, Message> {
-    SpinButton::new(label, on_change)
+pub struct SpinButton<'a, T, M>
+where
+    T: Add<Output = T> + Sub<Output = T> + PartialEq + PartialOrd + Display + Copy,
+{
+    /// The label that the spin button widget will have.
+    /// It is placed on the top of and centered on the spin button widget itself.
+    label: String,
```
mmstick, line 32: Should be `Cow<'a, str>`.

src/widget/spin_button/mod.rs
```diff
@@ -1,108 +1,276 @@
...
+                    .on_press((spin_btn.on_press)(decrement::<T>(
+                        spin_btn.step,
+                        spin_btn.value,
+                        spin_btn.min,
+                        spin_btn.max,
+                    )))
                     .into(),
-                text::title4(label)
+                // Using the title4 variant of text for consistency.
+                text::title4(format!("{}", spin_btn.value))
```
mmstick, line 197: Should be `label`, not the value. User must have control over formatting of the value.

bhh32:
Are they something I can help with fixing?

src/widget/spin_button/mod.rs
bhh32: I'm not sure I understand this ask here. Is the label in this case a text representation of the value? I took it as the title of what the spin button is for.  
  
So currently, as implemented, the label is on the top center of the widget. Whereas if it is supposed to be a formatted text representation of the current value, I misinterpreted the intent.  
  
If that's the case I'll definitely refactor to make it the right way. I may change the parameter name from label to something else for better clarity of what it's supposed to be though.

PR #724, opened 2024-11-19
feat!(spin_button): refactor and support vertical widget variant

mmstick:
@bhh32 Needs review. This is closer to the original behavior of the spin button. Fixed the button styling to use accent colors. Labels are now formatted in the example. Added overflow/underflow checks. And added `fraction::Decimal` spin button to example.


bhh32 (approved):
Yes, I believe this will work just fine. I was also working on the fixes that look very similar, but your style additions are a good touch. Thank you!

PR #728, opened 2024-11-21
Changed cosmic::command to cosmic::task, and made cosmic::Task consistent with cosmic::app::Task

Adam-Cosner:
This changes the cosmic::command module to cosmic::task ever since the change from Command to Task, and also fixes an inconsistency where cosmic::Task was a re-export of iced::Task while cosmic::app::Task was a different type alias

### DIFF
--- a/examples/about/src/main.rs
+++ b/examples/about/src/main.rs
@@ -61,7 +61,7 @@ impl cosmic::Application for App {
         &mut self.core
     }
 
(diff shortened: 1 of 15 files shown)


wash2 (approved):

wash2:
cosmic-comp uses libcosmic but it doesn't need the app module or the winit feature. I think all libcosmic apps should be using the winit feature, so is it a problem?

PR #737, opened 2024-11-26
Fix segmented button model index out of bounds when setting position

Tony4dev:
When trying to set position greater then length of the Deque, the method fallbacks to position equal to the Deque length. However, that value changes immediately after getting this fallbacked position with removal of the item (to prepare it for re-instert) and as a result the code panics with index out of bounds error.

Reordering of the code fixes this issue.

### DIFF
--- a/src/widget/segmented_button/model/mod.rs
+++ b/src/widget/segmented_button/model/mod.rs
@@ -355,9 +355,10 @@ where
             return None;
         };
 
+        self.order.remove(index as usize);
+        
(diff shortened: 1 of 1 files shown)


mmstick (dismissed):

CristianKerr:
@mmstick please one more approval, I had to fix formatting (extra spaces). Thanks

Tony4dev:
@mmstick please one more approval. I had to fix formatting (extra spaces), thanks.

mmstick (approved):

PR #739, opened 2024-11-27
[Toggler, List column] Padding/spacing fixes

git-f0x:
This removes the extra padding on the toggler after the rebase, which caused misalignment in apps (not quite sure why spacing adds padding).
screenshot-2024-11-27-23-24-13 screenshot-2024-11-27-23-26-14

Also fixes the header corner gaps in apps without a content container (e.g. `cosmic-term`).

Edit: The second commit matches the spacing/padding variables used in the list column to designs, and removes the `list::container` widget, since it's sparingly used (in 6 places across all `pop-os` repos), and its padding was often overriden when used.
A container with `.class(theme::Container::List)` can be used instead.
If this isn't desired, it can be added back, with possibly the padding removed.

### DIFF
--- a/examples/cosmic/src/window.rs
+++ b/examples/cosmic/src/window.rs
@@ -17,7 +17,7 @@ use cosmic::{
     prelude::*,
     theme::{self, Theme},
     widget::{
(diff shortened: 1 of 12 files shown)


mmstick (approved):

git-f0x:
I'll close it.

PR #768, opened 2024-12-30
[WIP] feat: TableView widget

Adam-Cosner:
Currently fully meets the base design standard, the following features are to be implemented in the future as needed: 

- [x] Context menus
    - [x] Item context menu
    - [x] Category context menu
- [ ] Drag n Drop
- [ ] Keyboard control
    - [ ] User-defined keyboard actions
- [ ] Custom widget instead of deferring to other widgets
    - [ ] in-widget control of compact and standard view switching


src/widget/table/mod.rs
```diff
@@ -0,0 +1,20 @@
...
+pub mod widget;
+pub use widget::TableView;
+
+pub type SingleSelectTableView<'a, Item, Category, Message> =
+    TableView<'a, SingleSelect, Item, Category, Message>;
+pub type SingleSelectModel<Item, Category> = Model<SingleSelect, Item, Category>;
+
+pub type MultiSelectTableView<'a, Item, Category, Message> =
+    TableView<'a, MultiSelect, Item, Category, Message>;
+pub type MultiSelectModel<Item, Category> = Model<MultiSelect, Item, Category>;
```
mmstick, line 20: Make sure to provide functions for these.

examples/table-view/src/main.rs
```diff
@@ -0,0 +1,210 @@
...
+            }
+        }
+        Task::none()
+    }
+
+    /// Creates a view after each update.
+    fn view(&self) -> Element<Self::Message> {
+        cosmic::widget::responsive(|size| {
+            if size.width < 600.0 {
+                table::SingleSelectTableView::new(&self.table_model)
```
mmstick, line 197: The standard practice would be to use a function instead of a new method.

```rs
table::single_select(&model)
```

src/widget/table/mod.rs
mmstick: I believe that you could do this by providing a `fn table()` function, where the model will decide if it's multi or single-select.

src/widget/table/model/mod.rs
```diff
@@ -0,0 +1,360 @@
+pub mod category;
```
mmstick, line 1: There's a lot of duplicated code here with the segmented button model. Perhaps we should make something more generic that is shared between them.

src/widget/table/mod.rs
Adam-Cosner: Just added the function

src/widget/table/model/mod.rs
Adam-Cosner: I think there is some code we could theoretically merge but they ultimately still need to be separate as they are fundamentally storing different data

src/widget/mod.rs
```diff
@@ -316,7 +316,7 @@ pub use spin_button::{spin_button, vertical as vertical_spin_button, SpinButton}
 pub mod tab_bar;
 
 pub mod table;
-pub use table::table;
+pub use table::{compact_table, table};
```
mmstick, line 319: Make sure to include the `#[doc(inline)]` line and check how it looks in the cargo docs.

Adam-Cosner: Just did, thanks for reminding me

mmstick:
@Adam-Cosner Need any help with this? We could merge as long as it's in a decent state as a MVP

mmstick:
Needs to be rebased. The merge pulled in a lot of unrelated commits

mmstick:
I can do the rebase from my end if you do not mind the help.

Adam-Cosner:
Yes please I don't know how to rebase haha

mmstick:
https://github.com/jj-vcs/jj

This utility might also be helpful.

Adam-Cosner:
Okay, that somehow made it worse

PR #771, opened 2024-12-31
Fixes for context menu for segmented button

jasonrhansen:
Fixes navbar context menu issue in Cosmic Files:
https://github.com/pop-os/cosmic-files/issues/381

It also makes it so clicking out of a menu with right and middle mouse buttons closes the context menu, just like left-clicking does.


mmstick (approved):
Thanks!

PR #773, opened 2025-01-03
TextInput: use Cow instead of reference for label, helper text and error text

sgued:
This is useful to be able to translate the label directly in the `view` function, instead of having to store a translation in the app state.


sgued:
I also notice that `error` is never used.

mmstick (approved):

PR #789, opened 2025-01-16
Small Calendar widget update

Tony4dev:
I Added Default and Clone interface to CalendarModel. Also I made the field 'visible' of the model public (like the 'selected' field) because we might want to use it to directly affect what the calendar shows. Added another QoL method to the model.


Tony4dev:
I see @edfloreshz added the interfaces in another PR, I am closing this one and open one withouth conflicts.

PR #797, opened 2025-01-22
chore: Update ron to 0.9

Drakulix:
Ron 0.9 has been in the works for years at this point (0.8 the latest stable was tagged in August 2023), but sadly has stalled quite a bit: https://github.com/ron-rs/ron/issues/496

It adds a bunch of really nice features, like support for tagged or untagged enum variants, but more importantly it has a new `RawValue` type (similar to `serde_json::RawValue`), which allows us to address a few shortcomings like: https://github.com/pop-os/cosmic-settings-daemon/pull/68

I say, lets bite the bullet and try the alpha-version of it. It seems remaining issues are only around `#[serde(flatten)`, which we don't use: https://github.com/search?q=org%3Apop-os%20%23%5Bserde(flatten)%5D&type=code


mmstick (approved):

HeitorAugustoLN:
RON 0.9.0 stable is now released!

PR #808, opened 2025-02-18
Tooltips and Better Surface Management

wash2:
This adds some improvements for popup, subsurface, and tooltip handling for wayland clients. It has the menu bar and menu tree changes generally removed for now until I get that working properly. I'll draft some updates for a few clients before marking this ready.


examples/application/src/main.rs
```diff
@@ -56,6 +85,23 @@ pub enum Message {
...
+            Message::Surface(s) => MessageWrapper::Surface(s),
+            m => MessageWrapper::Message(m),
+        }
+    }
+}
+
+impl From<SurfaceMessage> for Message {
+    fn from(value: SurfaceMessage) -> Self {
+        Message::Surface(value)
+    }
```
mmstick, line 104: Can this be handled transparently in libcosmic without boilerplate on the app developer side?

mmstick:
I updated libcosmic on this PR to this branch https://github.com/pop-os/cosmic-settings/pull/1016
Then I noticed that none of the dropdowns are working when using this branch.

examples/application/src/main.rs
wash2: Not that I'm aware of. I've implemented it such that it only needs to be implemented for apps that use this feature. Maybe a macro could work but I'm not sure.

wash2:
> I updated libcosmic on this PR to this branch [pop-os/cosmic-settings#1016](https://github.com/pop-os/cosmic-settings/pull/1016) Then I noticed that none of the dropdowns are working when using this branch.

They should work now after the latest commit. The new popup dropdowns involve some extra messages and configuration.

mmstick:
I've come up with a better solution that can eliminate the boilerplate. Need some time to improve it though.

wash2:
I'll work on updating some of the other apps with the improvements, and https://github.com/pop-os/iced/pull/215 is required for this too

examples/application/src/notif.rs
```diff
@@ -0,0 +1,89 @@
+use std::error::Error;
```
ids1024, line 1: This file seems to be an unrelated change, and the module isn't being used. So either it should actually be imported in `main.rs`, or removed, probably.

ids1024:
This changes the applet example to no longer use `view_window` for the applet popups. So we could do that in cosmic-applets, but https://github.com/pop-os/cosmic-applets/pull/826 doesn't change that yet? (Which is fine.)

And this should help with letting us make drop down menus in apps use proper `wl_popup`s, but also isn't changed so far? There's a popup or two in places like cosmic-settings which may be problematic in a small enough tiled window, so it would be nice to improve this, though it isn't the most important issue.

If the the only expected change to behavior so far is the tooltips (in the app-list, minimize, button applets), those seem to be working well. I wonder if the styling could be a little better in terms of the exact margins, and spacing from the applet, or such, but that can be tweaked later.

wash2:
> This changes the applet example to no longer use `view_window` for the applet popups. So we could do that in cosmic-applets, but [pop-os/cosmic-applets#826](https://github.com/pop-os/cosmic-applets/pull/826) doesn't change that yet? (Which is fine.)
> 
> And this should help with letting us make drop down menus in apps use proper `wl_popup`s, but also isn't changed so far? There's a popup or two in places like cosmic-settings which may be problematic in a small enough tiled window, so it would be nice to improve this, though it isn't the most important issue.
> 
> If the the only expected change to behavior so far is the tooltips (in the app-list, minimize, button applets), those seem to be working well. I wonder if the styling could be a little better in terms of the exact margins, and spacing from the applet, or such, but that can be tweaked later.

Ya, apps that want to use popups can now use them for dropdowns, but they currently have to opt into it by setting some of the parameters. The updates left that out, but we could start by updating the dropdowns in cosmic-settings that are likely problematic.

Drakulix:
> The updates left that out, but we could start by updating the dropdowns in cosmic-settings that are likely problematic.

We should in general update everything to use this, as otherwise the dropdowns might still be partially hidden behind other windows.

ids1024:
I see an error building with `cargo check --features tokio,wayland,winit`:

```
error[E0282]: type annotations needed for `widget::dropdown::widget::Dropdown<'_, S, Message, _>`
  --> src/widget/dropdown/mod.rs:45:9
   |
45 |     let dropdown = Dropdown::new(selections, selected, on_selected);
   |         ^^^^^^^^
   |
help: consider giving `dropdown` an explicit type, where the type for type parameter `AppMessage` is specified
   |
45 |     let dropdown: widget::dropdown::widget::Dropdown<'_, S, Message, AppMessage> = Dropdown::new(selections, selected, on_selected);
   |                 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
```

src/widget/wayland/tooltip/widget.rs
```diff
@@ -0,0 +1,678 @@
...
+                            let s = settings.clone();
+                            let view = view.clone();
+                            let bounds = layout.bounds();
+
+                            let sm = crate::surface::Action::Task(Arc::new(move || {
+                                let s = s.clone();
+                                let view = view.clone();
+                                let is_hovered = is_hovered.clone();
+                                Task::future(async move {
+                                    _ = tokio::time::sleep(delay).await;
```
ids1024, line 486: This line causes a build error without the `tokio` feature.

ids1024:
https://github.com/pop-os/cosmic-workspaces-epoch/pull/142 has updates for cosmic-workspaces. It doesn't really need popups or tooltips anywhere, but that confirms that the subsurface handling is still working properly anyway.

mmstick:
Ready to merge?

wash2:
> Ready to merge?

https://github.com/pop-os/iced/pull/215 is required first, but yes, I think so.

mmstick:
Merged

mmstick (approved):

PR #814, opened 2025-03-05
refactor!: improve surface action handling and adjust libcosmic API

mmstick:
For https://github.com/pop-os/libcosmic/pull/808

- Adjusts the surface message API to not require type conversions from the surface message type.
- Renames `surface_message` module to `surface`, and `SurfaceMessage` to `surface::Action`
- `surface::MessageWrapper` no longer necessary
- Moved surface action functions to `surface::action` module
- Renames `cosmic::app::Message` to `cosmic::Action` to reduce confusion, in line with how iced names runtime message types
- Likewise renames `cosmic::app::cosmic::Message` to `cosmic::app::Action`
- Moves commands used by libcosmic to `cosmic::command`
- Added some missing copyright headers

I've also experimented with removing the need for `Core` in the app struct but I'll leave that for a future PR.


mmstick:
I added Fn closures to the tooltip and responsive widgets as input parameters to their functions so that it isn't necessary to have an additional wrapper or `From/Into` type constraints. Which I've done in the cosmic-applets [here](https://github.com/pop-os/cosmic-applets/pull/839/commits/7a6a0b810d66b272facfaeb0d6fcc1f97258b2a7#diff-d3867156d53bc5c03cc97adafbdafd28135808886cd655919e2a658788bf4b96R1423).

Other than that, no behaviors have actually changed here. I've just renamed the `cosmic::app::Message` and `cosmic::app::cosmic::Message` types and moved them and their functions to better locations. It's always been confusing to see `cosmic::app::Message<Message>>` as a type signature. Such as `iced::Task<cosmic::app::Message<Message>>`, which `cosmic::app::Task<Message>` is just an alias for.

All application messages handled internally by libcosmic are in the [cosmic::app::Action](https://github.com/pop-os/libcosmic/blob/surface-improv/src/app/action.rs#L16) type, which is handled [here](https://github.com/pop-os/libcosmic/blob/surface-improv/src/app/cosmic.rs#L207). All of the `Application` methods that return tasks have been returning `iced::Task<cosmic::Action<Message>>` to allow applications to send messages to libcosmic's internal update method. I've been avoiding the need to require methods returning elements to use `cosmic::Action` as the message type, but there are a few that unfortunately do.

wash2:
Ok, looking at[ the applet updates](https://github.com/pop-os/libcosmic/blob/surface-improv/src/app/cosmic.rs#L207), it appears the tooltip view `Element` type still needs to be adjusted, because no tooltips are created by the buttons on hover. I made a small adjustment that should resolve that issue, and otherwise, it seems that the applet buttons panic on hover (during tooltip creation), because they do not forward the Surface message to be handled as the internal type. I will push my changes, so you can see what I mean.

wash2 (approved):

PR #817, opened 2025-03-07
fix: Support Spawning Processes on macOS

vkhitrin:
Allow using pipe on macOS when attempting to spawn a process.


vkhitrin:
I have encountered this during https://github.com/vkhitrin/cosmicding/issues/61.

Using this, I was able to recompile my application on the latest cosmic:
image

mmstick (approved):

PR #820, opened 2025-03-12
Visually align window corners to content

### LINKED ISSUE #767: Visual refinement for nested rounded corners
The nested rounded corners that appear in some places like the window and window sidebar corners currently use the same radius variable. Since there's a padding of 8 between them, this leads to the outer radius (window corner) to visually stick out:
screenshot-2024-12-30-16-04-52

It could be improved by increasing the outer radius by 1/2 of padding (adding the entire padding is mathematically correct for achieving concentric corners, but doesn't look good), which visually aligns the corners:
screenshot-2024-12-30-16-02-14

This would be implemented by having the outer radius use the following rule:
`window_corner_radius = radius_s().map(|radius| if radius < 4.0 { radius } else { radius + 4.0 })`
This also ensures that the Square style remains square, and is unaffected by this change.

This issue is created to gather feedback on the change.
CC @pop-os/ux

git-f0x:
This makes the window corners have a radius of 1/2 padding (4) larger than the content (context drawer, navbar), for values of `@radius_s` of 4 or more.
Also sets the progress bar radius to `@radius_xl` to match designs.

Closes #767.

### DIFF
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -838,6 +838,13 @@ impl<App: Application> ApplicationExt for App {
             content_col.into()
         };
 
+        // Ensures visually aligned radii for content and window corners
+        let window_corner_radius =
+            crate::theme::active()
+                .cosmic()
+                .radius_s()
+                .map(|x| if x < 4.0 { x } else { x + 4.0 });
+
         let view_column = crate::widget::column::with_capacity(2)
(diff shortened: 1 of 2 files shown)


git-f0x:
Should I run cargo fmt? (those places weren't touched by the PR)

PR #827, opened 2025-03-20
fix(text_input): only set read only if the text input is editable

wash2:


mmstick:
I am almost ready to merge substantial refactoring of the text input that will fix this and several other issues. Will get it finished tonight.

PR #833, opened 2025-03-27
fix(text_input): support numpad keys

### LINKED ISSUE #702: BUG: numeric keypad no longer writes keys anymore
I can write numbers in text inputs while holding `shift + &` for example, but the keypad no longer works (since some time now)

### LINKED ISSUE #740: Num pad do not work with input fields
Hi,
My num pad doesn't work with the input fields (e.g. text input). Trying to turn num pad off and on doesn't help. 
I use latest Fedora, so there might be some incompatibility with the system maybe?

Edit: just want to add that the keys behave always as their alternative functions like arrows, insert, pg up, pg down etc..

mmstick:
Compared with the current iced implementation of text_input, and made changes based on that to get better character input support.

Closes #702 
Closes #740

### DIFF
--- a/src/widget/text_input/input.rs
+++ b/src/widget/text_input/input.rs
@@ -1303,6 +1303,7 @@ pub fn layout<Message>(
     layout::Node::with_children(limits.resolve(width, size.height, size), nodes)
 }
 
+// TODO: Merge into widget method since iced has done the same.
 /// Processes an [`Event`] and updates the [`State`] of a [`TextInput`]
 /// accordingly.
 #[allow(clippy::too_many_arguments)]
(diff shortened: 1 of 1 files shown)


src/widget/text_input/input.rs
```diff
@@ -1579,7 +1580,18 @@ pub fn update<'a, Message: Clone + 'static>(
...
+            key,
+            modified_key,
+            text,
+            physical_key,
+            ..
+        }) => {
+            dbg!(&key);
+            dbg!(&modified_key);
+            dbg!(&physical_key);
+            dbg!(&text);
```
wash2, line 1593:
```suggestion
 ```

wash2 (changes requested):

wash2 (approved):

PR #843, opened 2025-04-08
Improve table sort

tafia:
While quickly reading at table widget, I figured the sort function was suboptimal.

This PR should speed it up.

- If the sort is already done, noop
- If the sort is on the same category but with a different direction, just reverse the order
  - **I believe this is something that happens a lot**
- If not, use `make_contiguous` to avoid reallocating new entities


mmstick (approved):

mmstick:
Merged. For some reason all my GitHub emails were sent to spam the last few days.

PR #847, opened 2025-04-11
fix: Calculate Icons For Header Bar Individually

vkhitrin:
Based on the existing non Linux macro, we are required to calculate
icons individually.


vkhitrin:
**Perhaps a better approach will be to modify the macro for non Linux operating systems.**

I have first encountered it when trying to recompile on macOS, https://github.com/vkhitrin/cosmicding/issues/86.

Using this fix, I was able to compile for both macOS:
screencapture 380
And my on Linux platform running COSMIC DE:

mmstick (approved):

PR #849, opened 2025-04-12
Fixing build-time error on Non-Linux targets

UchiWerfer:
fixed error on non-linux targets due to use of a non-literal expression getting passed to the concat!()-macro


git-f0x:
Duplicate of #847. Sorry for introducing the issue!

PR #856, opened 2025-04-17
About page - allow custom license url link

Tony4dev:
This will make it possible to set custom license url if license is not at spdx.org.
I intentionally added only the commented code line in the about example so that by default it creates the spdx.org license url, but also it is clear there is other option.


Tony4dev:
Can this be please merged in? @mmstick

Tony4dev:
@mmstick is this going to be accepted please?

mmstick (approved):

Issue #855, opened 2025-04-17
fail build on windows


ti-integra:
testing example table-view

cargo check
```
error[E0433]: failed to resolve: use of unresolved module or unlinked crate `fde`
  --> src\desktop.rs:13:24
   |
13 | impl IconSourceExt for fde::IconSource {
   |                        ^^^ use of unresolved module or unlinked crate `fde`

error: expected a literal
   --> src\widget\header_bar.rs:445:21
    |
445 | /                     if self.maximized {
446 | |                         "window-restore-symbolic"
447 | |                     } else {
448 | |                         "window-maximize-symbolic"
449 | |                     },
    | |_____________________^
    |
    = note: only literals (like `"foo"`, `-42` and `3.14`) can be passed to `concat!()`

error[E0433]: failed to resolve: use of unresolved module or unlinked crate `fde`
  --> src\desktop.rs:16:13
   |
16 |             fde::IconSource::Name(name) => crate::widget::icon::from_name(name.as_str())
   |             ^^^ use of unresolved module or unlinked crate `fde`
   |
   = help: if you wanted to use a crate named `fde`, use `cargo add fde` to add it to your `Cargo.toml`

error[E0433]: failed to resolve: use of unresolved module or unlinked crate `fde`
  --> src\desktop.rs:23:13
   |
23 |             fde::IconSource::Path(path) => {
   |             ^^^ use of unresolved module or unlinked crate `fde`
   |
   = help: if you wanted to use a crate named `fde`, use `cargo add fde` to add it to your `Cargo.toml`

```

in other examples

```
 Checking ashpd v0.9.2
error[E0432]: unresolved import `std::os::fd`
   --> C:\Users\daniel.inacio\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\ashpd-0.9.2\src\desktop\camera.rs:20:37
    |
20  | use std::{collections::HashMap, os::fd::OwnedFd};
    |                                     ^^ could not find `fd` in `os`
    |
note: found an item that was configured out
   --> C:\Users\daniel.inacio\.rustup\toolchains\stable-x86_64-pc-windows-msvc\lib/rustlib/src/rust\library\std\src\os\mod.rs:182:9
    |
182 | pub mod fd;
    |         ^^
note: the item is gated here
   --> C:\Users\daniel.inacio\.rustup\toolchains\stable-x86_64-pc-windows-msvc\lib/rustlib/src/rust\library\std\src\os\mod.rs:181:1
    |
[...]

PR #861, opened 2025-04-28
fix(855): Windows build (rebased/simplified 2025-06-23)

joshuamegnauth54:
Closes: #855 

The original issue mentions that several examples fail to build. They fail to build on Linux as well due to bit rot. I started fixing some of those issues until I realized that this is a larger project than I expected. :laughing: So, this PR mainly fixes the Windows build.

### DIFF
--- a/src/desktop.rs
+++ b/src/desktop.rs
@@ -10,6 +10,7 @@ pub trait IconSourceExt {
     fn as_cosmic_icon(&self) -> crate::widget::icon::Icon;
 }
 
+#[cfg(not(windows))]
 impl IconSourceExt for fde::IconSource {
     fn as_cosmic_icon(&self) -> crate::widget::icon::Icon {
         match self {
--- a/src/widget/header_bar.rs
+++ b/src/widget/header_bar.rs
@@ -454,14 +454,10 @@ impl<'a, Message: Clone + 'static> HeaderBar<'a, Message> {
 
                 #[cfg(not(target_os = "linux"))]
                 let icon = {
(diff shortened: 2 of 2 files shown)


ti-integra:
build error

```
error: invalid format string: expected `}`, found `$`
   --> src\widget\header_bar.rs:428:43
    |
428 |                         "../../res/icons/{$name}.svg"
    |                                          -^ expected `}` in format string
    |                                          |
    |                                          because of this opening brace
    
```
    
    
the error point to here 
```rs
    #[cfg(not(target_os = "linux"))]
                let icon = {
                    widget::icon::from_svg_bytes(include_bytes!(format!(
                        "../../res/icons/{$name}.svg"
                    )))
                    .symbolic(true)
                    .apply(widget::button::icon)
                    .padding(8)
                };
```

using this also in windows
```rs
                let icon = {
                    widget::icon::from_name($name)
                        .apply(widget::button::icon)
                        .padding(8)
                };
```
was able build but without icons

image

joshuamegnauth54:
This works now if you would like to use it before it's merged. The fix is a one-liner for the library, but I'll tackle some of the examples in another PR since it's a larger project.

tredeneo:
I have upgrade to last commit and build without errors, with icons

image

joshuamegnauth54:
I reverted the changes to the examples because they require more changes anyway. So, now this patch is very small and easier to merge.

~~The failing test is not due to this patch.~~ Issue fixed upstream so tests pass after rebasing.

mmstick (approved):

PR #865, opened 2025-05-01
add custom ssd height

cramt:
add option for pop-os/cosmic-epoch#275

### DIFF
--- a/cosmic-theme/src/model/theme.rs
+++ b/cosmic-theme/src/model/theme.rs
@@ -89,6 +89,8 @@ pub struct Theme {
     pub is_high_contrast: bool,
     /// cosmic-comp window gaps size (outer, inner)
     pub gaps: (u32, u32),
+    /// cosmic-comp height of server-side decorations
+    pub ssd_height: u32,
     /// cosmic-comp active hint window outline width
     pub active_hint: u32,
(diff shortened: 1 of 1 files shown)


mmstick:
There is already a header density config here: https://github.com/pop-os/libcosmic/blob/master/src/config/mod.rs#L98-L99

cramt:
> There is already a header size config here: [`master`/src/config/mod.rs#L98-L99](https://github.com/pop-os/libcosmic/blob/master/src/config/mod.rs#L98-L99)

yes but that doesnt solve the issue i linked of having no SSD

git-f0x:
`ssd_enabled` should be a compositor config option, since it can just not use the header bar if false.
The compositor sets SSD height on its own, and reducing SSD height would visually break window control buttons (it could only be reduced to 32 without visual bugs), so changes likely aren't needed here?

cramt:
@git-f0x is right — the compositor sets the SSD height itself, and shrinking it below 32px just breaks the window controls, so this doesn't belong in cosmic-theme. The actual goal (pop-os/cosmic-epoch#275) is served by a compositor-side toggle instead: pop-os/cosmic-comp#2666, which needs no libcosmic change. Closing.

PR #886, opened 2025-05-28
cleanup: Remove Size::Hardcoded from applet size variants

ryanabx:
The most adequate replacement for applets that want to be custom-sized is to use `PanelSize::Custom(n)`. This does lose the granularity between width and height though.

Might need discussion on whether people use width and height in community-made applets?


Ignavar:
Could you give an example of how we could use PanelSize::Custom(n) this i was currently trying to set the size of my applet while i was able to get the height where i wanted by setting the size of the elements inside the width would not change beyond a certain size no matter what i tried

PR #924, opened 2025-07-21
Add `dbus_connection()` method to `app::Application` trait

ids1024:
This allows an application that wants to expose a DBus protocol to use the same session DBus connection, and provide a protocol on the same owned bus name.

Without this, a Cosmic app using `single-instance` that wants to provide a DBus protocol must provide it on a bus name other than `APP_ID`.

### DIFF
--- a/src/app/action.rs
+++ b/src/app/action.rs
@@ -22,6 +22,8 @@ pub enum Action {
     Close,
     /// Closes or shows the context drawer.
     ContextDrawer(bool),
+    #[cfg(feature = "single-instance")]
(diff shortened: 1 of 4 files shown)


ids1024:
I also notice we use `org.freedesktop.DbusActivation` as the protocol name for what seems to match the `org.freedesktop.Application` protocol? Not sure if there's some reason it was done that way, but that doesn't seem correct. (If we wanted to namespace a custom version, it probably shouldn't be `org.freedesktop.`).

PR #926, opened 2025-07-22
context-menu: allow borrowed content

jackpot51:


jackpot51:
This was not used by cosmic-files but may still be useful for other apps

wash2 (approved):

Issue #913, opened 2025-07-03
[Feature request] Support SSD with a config option


nanowuu:
SSD support would make integration with other compositors a lot better. I propose a config option that, when toggled, disables the CSD and titlebar buttons and makes the headerbar widget less tall, to accomodate for SSD. This would make libcosmic apps integrate better with KDE for instance and support features like having titlebar buttons on the left. I don't know if Windows and Mac do SSD in the same way as Wayland but allowing for native titlebars on those platforms would fix issues like https://github.com/pop-os/libcosmic/issues/523#issuecomment-2258796991.

mmstick:
Windows and Mac use CSD so it's not out of place. Not sure what integration issues KDE has with apps using CSD. They just look the same way they do here and everywhere else. We can't alter the headerbar dimensions without breaking COSMIC apps that rely on those dimensions.

nanowuu:
Sorry, that was badly worded. By "SSD" I meant something more like "native window decorations" i.e. draw the OS's default, whether that be through SSD on Linux or CSD on Windows and Mac. By integration I just meant that libcosmic apps can look out of place, especially when you have other buttons in your titlebars such as the ones KDE allows for:

<img width="795" height="311" alt="Image" src="https://github.com/user-attachments/assets/d0ee6646-1803-45b4-9bd2-b80aa9fbb009" />
In this image I have the "keep above other windows" and "keep below other windows" buttons enabled as well as the regular close, minimise, maximise buttons. 
There is also still the issue of not having the window snapping menu in Windows and Mac as mentioned in https://github.com/pop-os/libcosmic/issues/523#issuecomment-2258796991
It's also useful to not have duplicate window decorations in a situation where you have disabled titlebars for maximised applications, which is how my system is configured. On KDE applications this reduces wasted space considerably:

<img width="1920" height="350" alt="Image" src="https://github.com/user-attachments/assets/e75d2986-2a3f-4e62-9366-39686f52f9cf" />
Whereas in COSMIC apps I see duplicate buttons (and menu bars, but I know that the Wayland global menu situation is finicky at the moment, and that's not the point of this issue)

<img width="1920" height="368" alt="Image" src="https://github.com/user-attachments/assets/cae7d02b-9db5-4077-8973-6c50344cab70" />
If the header bar's dimensions can't be altered that's also fine, due to the above points supporting SSD would still have benefits.

Issue #776, opened 2025-01-05
[Suggestion] Maybe have the headerbar fall back to being a menu toolbar when client decorations are disabled?


netraptor:
So, right now if you disable client decorations you end up with something like this:
![image](https://github.com/user-attachments/assets/30929510-6865-4c7f-885c-fa748290c96d)

Would it be desirable to have the headerbar turn into a menu toolbar (no window dragging, no window controls) if client decorations are disabled? I'm wondering if that might be acceptable as an option in situations where OS-native window decorations are desirable (such as was mentioned in issue #523).

As seen in the picture above, window corner behavior would also need to be changed to look better within server-side decorations.

Issue #487, opened 2024-06-03
Make window decorations clickable at the very window edge/corner


boredsquirrel:
According to [this GTK change request](https://gitlab.gnome.org/GNOME/gtk/-/issues/6710):

I think GTK4 and the pretty similar libcosmic are good to use. But the big issue I have is the window decorations, that are too big, and still need exact mouse pointing to use them.

This compared to the huge but completely easy to click decorations in Windows, or the smaller but similar (default, breeze) KDE Plasma window decorations, is worse.

Note that some Plasma decorations seem to also fail here.

It would be good to expand the clickbox of the decorations so that they
1. Expand from the "close" circle to the very window edge, making fullscreen windows easy to close by moving the cursor to the very edge of the screen (this does not apply to the GNOME-like layout)
2. Expand the clickboxes to that they go to the top of the window and directly next to each other, so sliding from the top corner to the left will smoothly allow to use the other buttons.

This reduces the accuracy needed to close windows and thus highly improves accessibility for many users.

PR #939, opened 2025-08-08
feat: add SSD support via config option

nanowuu:
This PR adds support for SSD in libcosmic via a config option. 
CSD is default on Linux, but will be disabled if the option in the toolkit config file or in `app::settings::Settings` are true.
On MacOS and Windows, SSD is default: see the last comment on #523. It will be disabled if the option in the toolkit config file or in `app::settings::Settings` are false.
(I know it's not technically SSD on Mac and Windows but it functions mostly the same)

Fixes: 
- #913 
- #776 
- #437 and #487 (when SSD is enabled on a window manager which supports this) 
- #421 (when SSD is enabled on a window manager which supports this)
- #523 - partially.

### DIFF
--- a/src/app/cosmic.rs
+++ b/src/app/cosmic.rs
@@ -391,7 +391,7 @@ where
     pub fn style(&self, theme: &Theme) -> iced_runtime::Appearance {
         if let Some(style) = self.app.style() {
             style
-        } else if self.app.core().window.is_maximized {
+        } else if self.app.core().window.is_maximized || !self.app.core().window.client_decorations {
             let theme = THEME.lock().unwrap();
             crate::style::iced::application::appearance(theme.borrow())
         } else {
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -75,7 +75,14 @@ pub(crate) fn iced_settings<App: Application>(
         window_settings.resize_border = border_size as u32;
         window_settings.resizable = true;
     }
-    window_settings.decorations = !settings.client_decorations;
+
+    core.window.client_decorations = if cfg!(any(target_os = "windows", target_os = "macos")) {
+        settings.client_decorations || crate::config::client_decorations()
+    } else {
+        settings.client_decorations && crate::config::client_decorations()
+    };
+    window_settings.decorations = !core.window.client_decorations;
+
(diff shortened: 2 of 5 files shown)


pyroraptor07:
Tested this on KDE, Windows, and MacOS with the cosmic app template (did have to update the iced submodule) and so far it seems to work pretty well. The only thing I would add (more because of [this bug](https://github.com/pop-os/libcosmic/issues/766) on MacOS really) is that moving the window using the cosmic header bar should probably be disabled when using SSD.

pyroraptor07:
Also, on MacOS, the cursor doesn't change for the window controls for some reason, but that probably should be looked at separately.

nanowuu:
@pyroraptor07 thanks for testing! I think that the window should still be moveable by the header bar because it means you don't have to aim for a small titlebar - this is how it functions on KDE and it's very useful. I would look at both of those Mac bugs, but I don't have access to one unfortunately.

pyroraptor07:
> @pyroraptor07 thanks for testing! I think that the window should still be moveable by the header bar because it means you don't have to aim for a small titlebar - this is how it functions on KDE and it's very useful. I would look at both of those Mac bugs, but I don't have access to one unfortunately.

That's fair enough, and those bugs should probably be looked at as separate issues anyways. I have access to a mac, but I have very little experience with the macos environment, so I'm not sure how much help I'd be troubleshooting them. But again, not related to this PR.

snoweuph:
shouldn't this be already managed by `cosmic::app::settings::Settings.client_decorations()` ?

image

---

I found this MR, after noticing that:

```rs
cosmic::app::settings::Settings.client_decorations(false)
```  

on latest main branch (rev `03d0171bbe36664b4749c2b70297a288f35df59f`) results in Client+Server Side Decorations at the same time:
image

nanowuu:
I do use that, but this PR is supposed to add a config option (for users) as well as a settings option (for developers) and figure it out depending on OS as described in the PR description. It also should fix that issue that you found.

snoweuph:
thanks for the info and PR :3

Milli189:
hopefully this pr will get merged soon!

PR #972, opened 2025-09-12
fix(dropdown): missing icon on windows

Oppzippy:


src/widget/dropdown/widget.rs
```diff
@@ -373,8 +373,14 @@ pub struct State {
 impl State {
     /// Creates a new [`State`] for a [`Dropdown`].
     pub fn new() -> Self {
+        #[cfg(target_os = "linux")]
+        let icon = icon::from_name("pan-down-symbolic").size(16).handle();
+        #[cfg(not(target_os = "linux"))]
+        let icon = icon::from_svg_bytes(include_bytes!("../../../res/icons/pan-down-symbolic.svg"))
```
mmstick, line 379: This should be a static global variable, and shared between other widgets that are also using this icon. The multi dropdown and table widgets also use this icon.

mmstick (changes requested):

PR #975, opened 2025-09-15
Create libcosmic.ftl PL initial translation

VandaLHJ:
Initial translation, i hope it's good enough.


FurkanAdmin:
Shouldnt this file be placed under pl/libcosmic or a similar path, rather than being added to the i18n root directory?

VandaLHJ:
Yes, it should be.

PR #980, opened 2025-09-24
feat(wayland): corner-radius protocol support

wash2:
requires https://github.com/pop-os/iced/pull/242


wash2:
The iced update also fixes #984

PR #996, opened 2025-10-02
Added ability for applications to propose their own preferred languages

UchiWerfer:
This gives applications that might have their own system of handling what language they want their UI to be displayed in a way of communicating this preference with libcosmic


UchiWerfer:
I closed this PR as I accidentally added an unrelated commit.

PR #1006, opened 2025-10-07
Added ability for applications to propose their own preferred languages

UchiWerfer:
This gives applications that might have their own system of handling what language they want their UI to be displayed in a way of communicating this preference with libcosmic.


UchiWerfer:
Opened this newly to still propose the change suggested in #996 that was closed due to an accidental mistake

PR #1020, opened 2025-10-13
Add helper for accumulating scroll into discrete delta

ids1024:
Improved version of scroll handling code from cosmic-workspaces and cosmic-applet-workspaces.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -108,6 +108,8 @@ pub mod task;
 
 pub mod theme;
 
+pub mod scroll;
+
 #[doc(inline)]
 pub use theme::{Theme, style};
 
--- a/src/scroll.rs
+++ b/src/scroll.rs
@@ -1,0 +1,112 @@
+use iced::Task;
+use iced::mouse::ScrollDelta;
+use std::time::{Duration, Instant};
+
+// Number of scroll pixels before changing workspace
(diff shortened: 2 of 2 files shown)


ids1024:
I think I've come up with a good way to handle a "rate limit" (I think that's a slightly better term than "debouncing" for what it's doing here). The rate limit is a configurable `Duration`. If set, it will produce at most one discrete event on a particular axis per that unit of time.

I think this works pretty well, but there are three values that can be tuned here easily if someone thinks scrolling is a bit to fast/slow/etc.:
- `SCROLL_PIXELS`: the number of pixels that is treated as a discrete change, for `ScrollDelta::Pixels` events
- `SCROLL_TIMEOUT`: the amount of time since the last scroll before old accumulated scroll events are discarded
- The configurable `rate_limit`: if the scroll events happen too quickly, it will produce a discrete scroll of 1 or -1 at this rate at most

This can be tested in https://github.com/pop-os/cosmic-workspaces-epoch/pull/220 and https://github.com/pop-os/cosmic-applets/pull/1102, which set the same `rate_limit`. We may want to experiment with different values of that if we want to use this for some other purposes. (Say if we wanted to use this for volume changes; though that could also be done in a non-discrete way, even if display rounded to an integer.)

wash2 (approved):

PR #1031, opened 2025-10-29
desktop: add CRX id extraction + matcher; prefer_svg only for -symbolic; tests

stephanbuys:
Fixes PWA icon resolution without browser hardcoding.

- Add public extract_crx_id() and CRX matcher (Exec --app-id, StartupWMClass=crx_<id>, Icon contains id)
- Prefer SVG only for -symbolic
- Unit tests for extraction/matching

Risk: low; no IO in resolver path. Logging policy unchanged.

### DIFF
--- a/src/desktop.rs
+++ b/src/desktop.rs
@@ -304,7 +304,12 @@ fn match_startup_wm_class(
 
 #[cfg(not(windows))]
 fn is_crx_id(candidate: &str) -> bool {
-    candidate.len() == 32 && candidate.chars().all(|c| matches!(c, 'a'..='p'))
+    is_crx_bytes(candidate.as_bytes())
+}
+
+#[cfg(not(windows))]
+fn is_crx_bytes(bytes: &[u8]) -> bool {
+    bytes.len() == 32 && bytes.iter().all(|b| matches!(b, b'a'..=b'p'))
 }
 
 #[cfg(not(windows))]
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -218,10 +218,16 @@ exclude = ["iced"]
...
 
+[dev-dependencies]
+tempfile = "3.13.0"
+
 
 [patch."https://github.com/pop-os/libcosmic"]
 libcosmic = { path = "./" }
 
+[patch."https://github.com/pop-os/cosmic-text.git"]
+cosmic-text = { path = "../cosmic-text" }
```
mmstick: This needs to be removed. Local path override.

src/desktop.rs
```diff
@@ -14,13 +14,21 @@ pub trait IconSourceExt {
...
-            fde::IconSource::Name(name) => crate::widget::icon::from_name(name.as_str())
-                .size(128)
+            fde::IconSource::Name(name) => {
+                // Prefer SVG only for symbolic icons; otherwise allow themed resolver
+                // to select optimal raster assets to avoid small-size SVG rendering issues.
+                let named = if name.ends_with("-symbolic") {
+                    crate::widget::icon::from_name(name.as_str()).prefer_svg(true)
+                } else {
+                    crate::widget::icon::from_name(name.as_str())
+                }
```
mmstick: icon::from_name is already checking if the name ends with `-symbolic`, and it sets `symbolic(true)` if so. Maybe we should simply set `prefer_svg`  based on that check too?

stephanbuys: Agreed, will refactor this.

mmstick:
Can you rebase on master? It's pulling in a conflicting git history. Either `git rebase origin/master` or switch to an updated master branch and `git cherry-pick`.

mmstick:
Thanks

mmstick:
Seems fine. We can improve it if there's issues found when testing it.

PR #1046, opened 2025-11-11
feat(icon): bundle icons for non-unix platforms & optimize

### LINKED ISSUE #225: Helper for bundling icons but allowing system themes to override them
After #224 we bundle two icons with libcosmic. There are several other icons used by libcosmic that could be bundled, thus allowing for sane fallbacks on systems without any freedesktop icon theme installed (macOS, Windows), or without cosmic-icons installed.

```
close-menu-symbolic
document-properties-symbolic
edit-clear-symbolic
edit-copy-symbolic
go-next-symbolic
go-previous-symbolic
list-add-symbolic
list-remove-symbolic
object-select-symbolic
open-menu-symbolic
pan-down-symbolic
system-lock-screen-symbolic
system-search-symbolic
window-close-symbolic
window-maximize-symbolic
window-minimize-symbolic
```

### LINKED ISSUE #701: Invisible window controls (close, minimize)
Hi,

When I'm running the examples, the buttons to close and minimize the window are invisible.
When hovering over the correct positions, they still get highlighted and they still work.

I have not installed the cosmic DE on my device - could this be a missing theme?
If so - is there a way for applications to come with an integrated default theme?

System Specs:

```
OS: Garuda Linux Bird of Prey x86_64
Kernel: 6.11.6-zen1-1-zen
Resolution: 2560x1440
DE: Plasma 6.2.3
WM: kwin
CPU: AMD Ryzen 7 5800X (16) @ 4.851GHz
GPU: NVIDIA GeForce RTX 2080 Ti
```

I'm interested in using libcosmic as my ui toolkit of choice. Is that an intended use case or should I stick to vanilla `iced` instead?

### LINKED ISSUE #763: [macOS] Icons From Builtin Widgets With Icons Using `widget::icon::from_name()` Do Not Render
While working to adapt my COSMIC based application (https://github.com/vkhitrin/cosmicding) to support macOS, I have noticed that most of the builtin widgets do not render icons.
As far as I am aware, this occurs due to macOS not having XDG (and cosmic-icons) icons available. 

Side note, `widget::button::link().trailing_icon(true)` is rendered.
(tested on commit https://github.com/pop-os/libcosmic?rev=953685a88267894ea3fa40f2b99139c3c4e784d6#953685a88267894ea3fa40f2b99139c3c4e784d6)

Example:
screencapture 3

I am not sure if supporting macOS is in the current scope of `libcosmic`, but I would like to raise the following questions:
1. For cross-platform systems which do not install `cosmic-icons` or XDG icons, should we expect the builtin default widgets to render images? For example, the trailing icons works for `widget::button::link` but checkboxes for `widget::menu::MenuItem::CheckBox` do not display the check mark.
2. For cross-platform development, should developers bundle their own SVG(s), and provide a custom class for widgets that refers to those SVG(s)?

mmstick:
- Adds all action icon SVGs from the cosmic icon theme for bundling.
- Embeds bundled icons on non-unix platforms with `phf`.
    - Eliminates need for platform-specific code when using `icon::from_name`.
- `Named` now creates an `svg::Handle` or `image::Handle` on conversion to an `icon::Handle`.
    - Removes the `Named` variant of `icon::Handle`'s `Data` type, and all the special case handling for it.
    - Improves icon performance if `icon::Handle` is cached in the app instead of created in the view.

Closes #225
Closes #701
Closes #763
Closes #972
Closes #1007


jackpot51:
Would it make sense at this point to have cosmic-icons be a submodule?

src/widget/icon/bundle.rs
```diff
@@ -0,0 +1,150 @@
...
+/// Get a bundled icon on non-unix platforms.
+#[cfg(not(unix))]
+pub fn get(icon: &str) -> Option<super::Data> {
+    super::bundle::BUNDLED_ICONS
+        .get(&self.name)
+        .map(|bytes| super::Data::Svg(crate::iced::widget::svg::Handle::from_memory(*bytes)))
+}
+
+#[cfg(not(unix))]
+static ICONS: phf::Map<&'static str, &'static [u8]> = phf::phf_map!(
```
jackpot51: I'd like to see some script to automatically generate this

jackpot51 (approved):

PR #1060, opened 2025-11-29
fix: compiling on windows requires cosmic-icons in project root

Oppzippy:
crabtime provides crabtime::WORKSPACE_PATH to refer to the CARGO_MANIFEST_DIR of the top level crate being built, which means when building libcosmic directly, crabtime::WORKSPACE_PATH will work, but when building it as a dependency of another crate, crabtime::WORKSPACE_PATH will no longer refer to the path to libcosmic.

I don't think there's a good workaround, since when in the context of crabtime, CARGO_MANIFEST_DIR refers to the path to the crate generated by crabtime rather than to libcosmic.

This replaces crabtime with a simple build.rs script that generates a file in OUT_DIR.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,9 +1,11 @@
 fn main() {
     println!("cargo::rerun-if-changed=build.rs");
 
+    #[cfg(not(unix))]
     generate_bundled_icons();
 }
 
+#[cfg(not(unix))]
 fn generate_bundled_icons() {
     println!("cargo::rerun-if-changed=cosmic-icons");
 


Oppzippy:
My bad, I should have mentioned why I didn't do this in the PR: e0e84c35c77c8c8748631b54108a76eb4fb87731 causes problems when cross compiling from Linux, since my understanding is in build.rs `#[cfg(not(unix))]` applies to the host machine (since that's where the build script is running) rather than the compilation target.

The build script can check the [`CARGO_CFG_UNIX` env var](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts) to avoid running the code, but I don't think compiling it can be avoided. There's no real reason I didn't to that.

Issue #1071, opened 2025-12-09
Doesn't compile for Rust Windows apps


bhh32:
An compile error is thrown where it doesn't understand the escape characters for icon packing in the build.rs file.

mmstick:
#1066 did not fix this?

bhh32:
> #1066 did not fix this?

No, I tried compiling a project on a Windows machine yesterday and it failed to compile until I added the patch you just merged in a local fork. Figured I'd push it upstream since I made it anyway.

bhh32:
#1072 did fix it. I can now compile on Windows from the git repo.

PR #1072, opened 2025-12-09
Bug Fix: Update Windows escape character

bhh32:
This single line change to the build.rs file just changes the escape character from the *nix standard `\` to the Windows standard `/`. This is in relation to issue #1071.


bhh32:
> xkbcommon should not be compiled on Windows

I agree, unsure why it did during a cross compilation test.

PR #1082, opened 2026-01-01
feat: add support for additional search paths for icons

Pandapip1:
WIP, untested

Blocked by https://github.com/oknozor/freedesktop-icons/pull/30

Blocking partial solution for https://github.com/pop-os/cosmic-applets/issues/380

Needed to support application-provided icon directories in cosmic-applet-status-area

### DIFF
--- a/src/applet/mod.rs
+++ b/src/applet/mod.rs
@@ -281,17 +281,38 @@ impl Context {
         .class(crate::theme::Button::AppletIcon)
     }
 
+    pub fn get_icon_button_icon_with_path<'a, Message: Clone + 'static>(
+        &self,
+        icon_name: &'a str,
+        extra_search_path: Option<Vec<std::path::PathBuf>>,
+    ) -> Named {
+        let suggested_size = self.suggested_size(true);
+        widget::icon::from_name_with_path(icon_name, extra_search_path)
+            .symbolic(true)
+            .size(suggested_size.0)
+            .into()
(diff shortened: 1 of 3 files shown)


mmstick:
We should not be circumventing the icon searching functionality provided by https://github.com/pop-os/freedesktop-icons. If the cosmic-freedesktop-icons crate is unable to find an icon, we should determine why that is so and fix it.

Pandapip1:
> We should not be circumventing the icon searching functionality provided by https://github.com/pop-os/freedesktop-icons. If the cosmic-freedesktop-icons crate is unable to find an icon, we should determine why that is so and fix it.

This isn't? It's explicitly adding support for https://github.com/oknozor/freedesktop-icons/pull/30, which is needed to implement a correct implementation for `resolve_icon_in_directory` (https://github.com/pop-os/cosmic-applets/pull/1230) (note: I'll actually be removing that function in the PR, as it would be made redundant by `from_name_with_path`).

This can't be fixed by permanently adding a directory to the search path, as ~~the FD.O SNI spec calls for a one-off extra directory to be searched~~ seemingly this `IconThemePath` is a thing supported by GNOME and KDE SNI impls, but not technically in the spec. It's used by a small but notable (e.g. Dropbox) subset of applications though so it's something that needs to be supported. Ergo the upstream PR I submitted and this draft PR.

PR #1099, opened 2026-01-19
Menu item width

bhh32:
## Summary

Refactors `MenuItem` from an enum to a struct wrapper, adding a `.width()` builder method that can be applied to any menu item type. This lets applications control submenu widths without hardcoding values into libcosmic forks/patches.

## Problem

Currently, there's no way to set a custom width on menu items. This causes issues when menu content (like file paths) is longer than the default width, resulting is bad looking UI.

**Before**:
before_cosmic-edit_menu_ui

**After**:
after_cosmic-viewer_ui

## Changes

- Renamed `MenuItem` enum to `MenuItemKind`
- Created new `MenuItem` struct wrapping `MenuItemKind` with optional `width` field
- Added `.width(u16)` builder method
- Added convenience constructors: `button()`, `button_disabled()`, `checkbox()`, `folder()`, and `divider()`
- Updated `menu_items()` to apply width to all item types.
- Updated `responsive_menu_bar.rs` to use new API

## Usage

```rust
// Old API
menu::Item::Folder(fl!("recent-folders"), items)

// New API
menu::Item::folder(fl!("recent-folders"), items).width(300)
```

## Migration

Existing code using enum variants (Item::Button(...), Item::Folder(...), etc.) should switch to the method constructors (Item::button(...), Item::folder(...), etc.). The From<MenuItemKind> impl provides backwards compatibility for code that builds MenuItemKind directly.

### DIFF
--- a/examples/application/src/main.rs
+++ b/examples/application/src/main.rs
@@ -236,26 +236,26 @@ impl cosmic::Application for App {
                 (
                     "hi 1".into(),
                     vec![
-                        menu::Item::Button("hi 12", None, Action::Hi),
-                        menu::Item::Button("hi 13", None, Action::Hi2),
+                        menu::Item::button("hi 12", None, Action::Hi),
+                        menu::Item::button("hi 13", None, Action::Hi2),
                     ],
                 ),
                 (
                     "hi 2".into(),
                     vec![
-                        menu::Item::Button("hi 21", None, Action::Hi),
-                        menu::Item::Button("hi 22", None, Action::Hi2),
-                        menu::Item::Folder(
+                        menu::Item::button("hi 21", None, Action::Hi),
+                        menu::Item::button("hi 22", None, Action::Hi2),
+                        menu::Item::folder(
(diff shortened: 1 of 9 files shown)


mmstick:
It would be better if we calculate the width in the widget so that they're automatically resized.

mmstick:
This should also update the menu examples, and have examples included in the code docs

git-f0x:
Might also be nice if there was a context menu widget that has the same behavior as the menu bar popups, so that it can have nested menus (e.g. Open with... > list of apps in a nested menu in COSMIC Files). Or maybe the same widget and it just differs in the way it's spawned?

bhh32:
> This should also update the menu examples, and have examples included in the code docs

I can definitely update the examples and code docs.

bhh32:
@mmstick, I've updated the examples and doc comments. I also added min_width and max_width builder options, updated `MenuTree`, and `get_children_layout` to ensure that internal auto-resizing occurs and is clamped to the min_width and max_width restrictions.

PR #1134, opened 2026-02-19
fix: right-clicking any sidebar item makes all sidebar items bold

mariinkys:

Hi, recently I implemented a context menu in the navabar inside my app and I stumbled upon a bug that made all items bold when opening the context menu (in the navbar), I found the issue has already been reported in: https://github.com/pop-os/cosmic-files/issues/1154. This commit fixes the issue with I believe no side effects.

When a context menu is open, state.show_context.is_some() is true for every button, not just the one that was right clicked. So every button gets assigned font_hovered (which is semibold), making them all appear bold.
I've fixed it by checking whether the context menu belongs to this specific button, not just whether any context menu is open at all.

### DIFF
--- a/src/widget/segmented_button/widget.rs
+++ b/src/widget/segmented_button/widget.rs
@@ -246,7 +246,7 @@ where
         if let Some(text) = self.model.text.get(key) {
             let font = if self.button_is_focused(state, key) {
                 self.font_active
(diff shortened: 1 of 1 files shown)


jacobgkau:
Just a heads-up, this did fix the rest of the items going bold, but it didn't fix the right-clicked item remaining bold after the context menu is closed.

PR #1140, opened 2026-02-23
fix: ellipsize headerbar title instead of wrapping

hojjatabdollahi:
Before:
image

After:
image

____

### DIFF
--- a/src/widget/header_bar.rs
+++ b/src/widget/header_bar.rs
@@ -445,6 +445,10 @@ impl<'a, Message: Clone + 'static> HeaderBar<'a, Message> {
         std::mem::swap(&mut title, &mut self.title);
 
         widget::text::heading(title)
+            .wrapping(iced_core::text::Wrapping::None)
+            .ellipsize(iced_core::text::Ellipsize::End(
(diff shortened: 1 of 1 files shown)


git-f0x:
Just a note that on e.g. KDE, the titles are ellipsized in the middle, so maybe it makes sense to do the same in libcosmic.

hojjatabdollahi:
Interesting. I assume I should do the same for tab_text too. Right?

git-f0x:
Tab text might be different (at least for comp), since it often just ends with the name of the app, which isn't too useful.
I added ellipsizing at the end for the App Tray, since the info at the start is more important than seeing e.g. `xyz...(Fi)les`.

hojjatabdollahi:
Maybe the ui/ux team can chime in.

I don't think ellipsize::middle makes a lot of sense, it's not that readable.

image

hojjatabdollahi:
We can probably move forward with this PR, now that [Maria has Okayed it](https://chat.pop-os.org/pop-os/pl/cb7ap5tcbpba8881n7znj53dmr).

PR #1172, opened 2026-03-15
Feat(cosmic-config): '::new_data'

DanielKaleby:

It works like `::new_state`, but stores files in `~/.local/share/cosmic/com.system76.Example/` instead of `~/.local/state/cosmic/com.system76.Example/` (on Linux, idk about `dirs` crate in other system)

### DIFF
--- a/cosmic-config/src/lib.rs
+++ b/cosmic-config/src/lib.rs
@@ -54,6 +54,22 @@ fn get_state_dir() -> Option<PathBuf> {
     dirs::state_dir()
 }
 
+/// Get the data directory, with Flatpak sandbox support.
+fn get_data_dir() -> Option<PathBuf> {
+    // Check if we're running in Flatpak
+    if env::var_os("FLATPAK_ID").is_some() {
+        // Try HOST_XDG_DATA_HOME first
+        if let Some(host_data) = env::var_os("HOST_XDG_DATA_HOME") {
+            return Some(PathBuf::from(host_data));
+        }
+        // Fallback: try to construct from HOME
+        if let Some(home) = env::var_os("HOME") {
+            return Some(PathBuf::from(home).join(".local").join("share"));
+        }
+    }
+    dirs::data_dir()
+}
+
 #[cfg(feature = "subscription")]
 mod subscription;
 #[cfg(feature = "subscription")]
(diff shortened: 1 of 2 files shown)


DanielKaleby:
I guess this might not be a thing for some reason, right? Or is it okay to have other data in the config lib (as `state` already does)?

DanielKaleby:
I don't test it as flatpak, *just* with `$ just run config`.

DanielKaleby:
Hello @mmstick, could you please review this PR? I'd appreciate your feedback. Thanks!

mmstick (approved):
Looks OK. Works the same as new_state.

jackpot51:
@DanielKaleby what are you using this for? There are issues with portability because on macOS and Windows, `dirs::config_dir` and `dirs::data_dir` return the same directory.

DanielKaleby:
> @DanielKaleby what are you using this for? There are issues with portability because on macOS and Windows, `dirs::config_dir` and `dirs::data_dir` return the same directory.

Yes, I use this to store the solve history of a speedcubing stopwatch. Before this, we stored the history in the config folder, which is weird. I've never tested it on macOS, however.

mmstick:
I think it would be better if apps use their own data directory over the `cosmic` config system for this. App data files aren't configuration files so there is no benefit in using the cosmic-config's versioned RON config system.

> Yes, I use this to store the solve history of a speedcubing stopwatch

This may be a better fit for cosmic-config's state directory rather than data.

jackpot51:
Yes, we are going to be removing this due to the portability issues. I'd recommend to use state instead or not use cosmic-config.

DanielKaleby:
> on macOS and Windows, `dirs::config_dir` and `dirs::data_dir` return the same directory.

What's the real problem here? Isn't this supposed to point to different directories on different operating systems, and isn't there an exact correspondence between them all?

DanielKaleby:
I'll look for alternatives. @jackpot51 @mmstick , thanks for your work, thanks to this, I'm finally learning a little about Rust.

jackpot51:
On macOS and Windows, dirs::config_dir and dirs::data_dir point to the same place. This means you could never use config and data with different data structures at the same time.

PR #1196, opened 2026-03-26
Theme v2

wash2:
This should update the version of the theme in a way that won't lose existing custom themes, though apps should probably be updated before cosmic-settings. A future v3 of the theme could drop the `Rgba` and `Rgb` variants from ColorRepr. This seems to work ok with settings and files.


mmstick (dismissed):

jackpot51 (dismissed):

mmstick (dismissed):

git-f0x:
Would it maybe make sense to move `Roundness` from COSMIC Settings to libcosmic, similar to how `Spacing` already is?

wash2:
> Would it maybe make sense to move `Roundness` from COSMIC Settings to libcosmic, similar to how `Spacing` already is?

ya, that could make sense.

git-f0x:
For the header bar container gaps, they're between the window border from the wrapping container and the header bar corners (very small, but looks like artifacting in some cases, e.g. if the content behind it is bright for a dark window). That part minimized them by adapting the header corner radius for non-content container windows (like cosmic-term).
The gaps are there even then (but very hard to notice), because the border isn't a part of the same container.

Though yeah, that does seem to cause double-coloring, since the base header bar is only transparent when content container is true.
Maybe that part can be kept, and instead the HeaderBar container should always be transparent (though cosmic-comp SSDs need to be taken into account, so it should only unconditionally be set to true in `app/mod.rs`, rather than the transparent field being removed), and then the container with adapted corner radius can fill the background?

git-f0x:
https://github.com/pop-os/libcosmic/pull/1262

This should fix the double-coloring of non-content container header bars. The header container should be added back for better-looking corners for those windows (revert https://github.com/pop-os/libcosmic/pull/1196/changes/35093f9a0266edcff3eec41d98c2158b3baa2570).
If using compositor clipping, the issue might not be as visible.

Without the container, and with the container:
Screenshot_20260422_225431
Screenshot_20260422_225633

git-f0x:
Just a note that 8035ef9535fa87deb05d88594a6b6a412324c039 should be reverted, since header bars of non-content container apps will be transparent otherwise (#1262 prevented double-coloring).

mmstick (dismissed):

src/widget/menu/menu_inner.rs
```diff
@@ -818,13 +818,26 @@ impl<'b, Message: Clone + 'static> Menu<'b, Message> {
                                     .children()
                                     .nth(active.saturating_sub(start_index))
                             {
+                                let i = active.saturating_sub(start_index);
+                                let mut rad = styling.menu_border_radius;
+                                let rad_0 = theme.cosmic().radius_0();
+                                if start_index == end_index {
+                                } else if 0 == i {
```
git-f0x: This does nothing if `start_index == end_index`. Is that intended (edit: makes sense logically, though feels potentially confusing)?

Maybe a function similar to https://github.com/pop-os/libcosmic/blob/18a7a75e3316b1f4855777af677d989467f2da74/src/widget/list/list_column.rs#L245-L252 would be useful here?

mmstick (dismissed):

mmstick (approved):

hojjatabdollahi:
@wash2 when I run `cargo run -p multi-window` one is transparent one isn't. Pretty sure it's a bug.

wash2:
> @wash2 when I run `cargo run -p multi-window` one is transparent one isn't. Pretty sure it's a bug.

Hmm, it looks like both are blurred when I tested it, though one is more opaque, and there seems to be some flickering along with some other issues that might just be related to the example. I'll look more into it though, and hopefully get a fix out soon, so this example is in better shape.

hojjatabdollahi:
Yeah. It seems we're not tracking the child window, so changes to the theme, live settings,... isn't going to the child.

I'm having a different issue with it. There is no way to get the window id for the child window. So opening a context menu in the child opens it in the parent. 

(Pretty sure the same bug exists already, for example a drop down in the child will open in the parent window)

wash2:
> Yeah. It seems we're not tracking the child window, so changes to the theme, live settings,... isn't going to the child.
> 
> I'm having a different issue with it. There is no way to get the window id for the child window. So opening a context menu in the child opens it in the parent.
> 
> (Pretty sure the same bug exists already, for example a drop down in the child will open in the parent window)

To track the windows, it helps to create them using the libcosmic helper methods that create surface actions, and I should mention that LiveSettings is a per-surface override, and without setting it, libcosmic automatically applies defaults depending on the theme. If the iced tasks are used instead, libcosmic is bypassed, and so we have a harder time tracking the surface and deciding what its attributes should be.

PR #1221, opened 2026-04-04
[ListColumn] Add ListButton list items for special handling of button-like elements

git-f0x:
Adds an `add_button` method which can be used to add any element with an associated on_press/toggle-like action, with hover highlights.
I've removed dividers here, since they need to be full-width with the highlights, and it looks cluttered then. They are replaced by having small gaps between children, which I think looks cleaner and might work better with frosted glass.
But it can be changed easily if desired (though adds complexity in having to deal with double-coloring, which makes things not have the expected transparency with frosted glass).

[Screencast_20260404_170022.webm](https://github.com/user-attachments/assets/564744c2-5deb-45cc-b36c-a5d937d60628)

Here's how it looks like in COSMIC Store, since that uses custom styling which I removed here (since the column isn't wrapped in a container anymore):
image

Any tips as to if the API can be improved in some way are welcome, since it does feel a bit janky when doing it with toggles/checkboxes, since the Message needs to pass the inverse of the current state.
E.g. the below, ~~and the toggler doesn't animate unless clicked directly~~:
```
.add_button(
    settings::item::builder(&descriptions[label_keys["window_hint_toggle"]])
        .toggler(
            theme_manager.custom_window_hint().is_none(),
            Message::UseDefaultWindowHint,
        ),
    Message::UseDefaultWindowHint(theme_manager.custom_window_hint().is_some()),
)
```
Maybe a purely visual toggler method for the builder (would need further changes to toggler in order to animate), so it's not needed to pass the Message in 2 places?

Should a `selected` field be added for `add_button` (or a separate method) for buttons that use it (e.g. the user list in Settings)?

My COSMIC Settings fork can be used for testing, if needed: https://github.com/git-f0x/cosmic-settings


mmstick (dismissed):

mmstick:
Needs UX approval for the divider change

git-f0x:
Is having a method for a selected button something that would be good to add, since the user page in Settings currently uses that?
E.g. `add_button_selected(element, on_press, selected)`

Also, the button highlights in the default light theme are almost imperceptible (the highlight is light, instead of dark), so if someone with more COSMIC theme experience knows why that is, that would be great.

mmstick:
Perhaps a builder would be better

```rs
list::item::button(element)
    .on_press(message_fn)
    .selected(bool)
```

git-f0x:
I think this is quite nice now.
Adding buttons is done with e.g. `.add_button(list::button(element).on_press(Msg).selected(selected))`.
I toyed around with having just `add`, but that ends up needing `.into()` everywhere, and multiple `From` impls.

mmstick (dismissed):

mmstick:
Code-wise it's good but needs approval from @maria-komarova for the change to dividers.

git-f0x:
Just a note that it might be better to wait on merging this, since I won't have access to a computer until the middle of next week, so I won't be able to follow-up with PRs for apps adapted to this.

maria-komarova:
To make sure some chat conversations are reflected here. We should stick with the dividers stretching to the edges of the list view.
Another question is about the hover state. There is a general practice to reduce motion in the operating systems. And having pretty large areas flash switching to hover might be a bit distracting visually and can increase the cognitive load. That said having the whole item be clickable and having a pressed state for the items is a good improvement.

git-f0x:
This matches the discussed designs now. The wrapping container is back (and the style method), alongside full-width dividers.
The list buttons now only have a highlight on press or when selected.

If someone has an idea if toggler-like elements could be made to feel less jank with this, that would be great:
```
list::button(
    settings::item::builder(fl!("enable-export"))
        .description(fl!("enable-export", "desc"))
        .toggler(self.icon_global, |b| {
            Message::DrawerIcon(IconMessage::ApplyThemeGlobal(b))
        })
)
.on_press(Message::DrawerIcon(IconMessage::ApplyThemeGlobal(
    !self.icon_global
)))
```

git-f0x:
I changed the toggler settings item builder to return a ListButton, which avoids the jank. It also makes it so that no changes are needed in downstream code to get the new behavior for togglers.
I've also added a checkbox builder, since it's the same as togglers in terms of jank otherwise.
Radio widgets seem a bit difficult since they require a label, so I didn't make a builder method for that.

Also added a trait to convert provided widgets to the appropriate ListItem type, so the `add_button` method isn't needed anymore.

maria-komarova (approved):

mmstick (approved):

PR #1264, opened 2026-04-23
fix: make menu bar background respect dynamic round settings

### LINKED ISSUE #1260: Right click menus on the panel apps don't respect the "Round" settings
As the title says:

This is with slightly round:

Image

This is the round option:

Image

as you can see from the border of the right click menu it does change and becomes round, but the background still remains like the "slightly round" options and overlaps the rounded corners

Rakshat28:
Updated menu_border_radius in menu_bar.rs to use radius_m, matching standard dropdowns and correctly scaling with the dynamic Wayland popup border.

Fixes #1260

### DIFF
--- a/src/theme/style/menu_bar.rs
+++ b/src/theme/style/menu_bar.rs
@@ -71,7 +71,7 @@ impl StyleSheet for Theme {
                 background: component.base.into(),
                 border_width: 1.0,
                 bar_border_radius: cosmic.corner_radii.radius_xl,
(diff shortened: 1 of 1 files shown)


mmstick (approved):

git-f0x:
This is not the right fix. This breaks menu appearance, while applets are using popup_container and not anything from menu_bar. @mmstick
image

I think the applet context menu might just need `.map(|x| x + 1.0)`

PR #1276, opened 2026-05-04
Fix shortcut matching for non-Latin keyboard layouts

Stolenskiy:
- [x] I have disclosed use of any AI generated code in my commit messages. - If you are using an LLM, and do not fully understand the changes it is making to the code base, do not create a PR. - In our experience, AI generated code often results in overly complex code that lacks enough context for a proper fix or feature inclusion. This results in considerably longer code reviews. Due to this, AI authored or partially authored PRs may be closed without comment.

## Summary

Adds layout-independent physical-key shortcut matching support to `KeyBind`.

This allows COSMIC applications to match standard editing shortcuts by physical
key position instead of only by the layout-dependent logical `Key::Character(...)`
value.

## Problem

Shortcut matching currently relies on logical key values such as
`Key::Character("c")` or `Key::Character("v")`.

Under non-Latin keyboard layouts, the same physical key can produce a different
logical character. For example, with a Ukrainian/Cyrillic layout, the physical
`C` key no longer produces the Latin `"С"` (They look similar, but the codes are different)character. As a result, applications
that define standard editing shortcuts with Latin letter keys may fail to match
those shortcuts when a non-Latin layout is active.

This affects common shortcuts such as:

- `Ctrl+C`
- `Ctrl+V`
- `Ctrl+X`
- `Ctrl+A`
- terminal-specific shortcuts such as `Ctrl+Shift+C` and `Ctrl+Shift+V`

## Fix

This change adds `KeyBind::matches_physical(...)`, which matches shortcuts by
layout-independent physical key position (`Physical::Code(Code::Key*)`) while
preserving the existing logical matcher.

The existing `KeyBind::matches(...)` behavior remains unchanged.

This lets applications opt into physical-key matching only where appropriate,
for example for standard editing shortcuts, while keeping logical matching for
layout-sensitive shortcuts such as punctuation.

## Why both logical and physical matching are needed

Desktop applications need both shortcut matching modes:

- logical matching for shortcuts that should follow the active keyboard layout;
- physical matching for standard editing shortcuts that should stay on the same
  physical keys across layouts.

This PR provides the missing shared primitive in `libcosmic` without changing
existing shortcut behavior for current consumers.

## Related issues

Addresses:

- pop-os/cosmic-epoch#3264

Related:

- pop-os/cosmic-epoch#3465 — duplicate of pop-os/cosmic-epoch#3264

## Follow-up

This change provides the shared `libcosmic` support needed to fix the issue in
COSMIC applications.

Follow-up changes are still needed in downstream applications such as
`cosmic-term` and `cosmic-edit` to opt into `matches_physical(...)` for their
standard editing shortcuts.

### DIFF
--- a/src/widget/menu/key_bind.rs
+++ b/src/widget/menu/key_bind.rs
@@ -1,4 +1,4 @@
-use iced_core::keyboard::{Key, Modifiers};
+use iced_core::keyboard::{Key, Modifiers, key::Code, key::Physical};
 use std::fmt;
 
 /// Represents the modifier keys on a keyboard.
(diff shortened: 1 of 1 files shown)


jackpot51:
The PR template is mandatory

Stolenskiy:
> Шаблон PR є обов'язковим

Thanks, I updated the PR description to include the mandatory template checklist.
I also added the AI assistance disclosure to the commit message where applicable.

mmstick:
Same as #1144

PR #1284, opened 2026-05-17
fix(process): safely exit forked child after spawning command

xDAndrew:
---

## Summary

Fixed process handling in `spawn()` to prevent `cosmic-app-library` from becoming a zombie process after launching applications.

## Before

When launching any application through `cosmic-app-library`, the `cosmic-app-library` process could become a zombie process.

The parent process was waiting using `rustix::process::wait(...)`, while the forked child process was exiting via `std::process::exit(0)`. This could affect the parent process state and leave the launcher in a broken state.

## How to reproduce

1. Open `cosmic-app-library`.
2. Launch any application from it.
3. Open `cosmic-app-library`.

Expected behavior:
`cosmic-app-library` responsive after spawning applications;

Current behavior:
`cosmic-app-library` became zombie process and doesn't response until restart;

## After

The parent process now waits for the forked child using `libc::waitpid(...)` inside `tokio::task::spawn_blocking(...)`, avoiding blocking the async runtime.

The forked child now exits using `libc::_exit(...)` after spawning the target command and writing the child PID to the pipe.

As a result, the child process terminates correctly and `cosmic-app-library` continues working normally after launching applications.

## Impact

- Prevents zombie processes after launching applications from `cosmic-app-library`
- Keeps `cosmic-app-library` responsive after spawning applications
- Avoids blocking the async runtime while waiting for the forked process
- Improves process cleanup and launcher stability

---

## Note

I marked several issues as possibly related to this fix because the problem stopped reproducing on my local setup after applying the change.

I have been testing the current version locally for the last 3 days, and the issue has not reproduced so far.

I do not have solid proof that these issues are definitely fixed by this change, but I can no longer reproduce them locally.

### DIFF
--- a/src/process.rs
+++ b/src/process.rs
@@ -5,7 +5,7 @@
 use smol::io::AsyncReadExt;
 use std::io;
 use std::os::fd::OwnedFd;
-use std::process::{Command, Stdio, exit};
+use std::process::{Command, Stdio};
 #[cfg(feature = "tokio")]
 use tokio::io::AsyncReadExt;
 
@@ -56,12 +56,18 @@ pub async fn spawn(mut command: Command) -> Option<u32> {
 
     match unsafe { libc::fork() } {
         // Parent process
-        1.. => {
+        fork_pid @ 1..=i32::MAX => {
             // Drop copy of write end, then read PID from pipe
             drop(write);
             let pid = read_from_pipe(read).await;
             // wait to prevent zombie
-            _ = rustix::process::wait(rustix::process::WaitOptions::empty());
+            let _ = tokio::task::spawn_blocking(move || {
+                let mut status = 0;
+                unsafe {
+                    libc::waitpid(fork_pid, &mut status, 0);
+                }
+            }).await;
+
             pid
         }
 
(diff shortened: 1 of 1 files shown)


mmstick:
This would require a PR with this branch for review.

mmstick:
I'm rewriting the summary for brevity and removing the issues because it links to unrelated issues and this should not close any issues in other repositories.

src/process.rs
```diff
@@ -56,12 +56,18 @@ pub async fn spawn(mut command: Command) -> Option<u32> {
...
+        fork_pid @ 1..=i32::MAX => {
             // Drop copy of write end, then read PID from pipe
             drop(write);
             let pid = read_from_pipe(read).await;
             // wait to prevent zombie
-            _ = rustix::process::wait(rustix::process::WaitOptions::empty());
+            let _ = tokio::task::spawn_blocking(move || {
+                let mut status = 0;
+                unsafe {
+                    libc::waitpid(fork_pid, &mut status, 0);
```
mmstick, line 67: This should use rustix instead of libc

src/process.rs
```diff
@@ -56,12 +56,18 @@ pub async fn spawn(mut command: Command) -> Option<u32> {
...
     match unsafe { libc::fork() } {
         // Parent process
-        1.. => {
+        fork_pid @ 1..=i32::MAX => {
             // Drop copy of write end, then read PID from pipe
             drop(write);
             let pid = read_from_pipe(read).await;
             // wait to prevent zombie
-            _ = rustix::process::wait(rustix::process::WaitOptions::empty());
+            let _ = tokio::task::spawn_blocking(move || {
```
mmstick, line 64: The overhead of this shouldn't be needed since the fork only lives long enough to write the PID of the child it spawns.

mmstick:
Since it's a small change, I can make PRs based on this for quicker turnaround.

PR #1288, opened 2026-05-20
Selectable Text + Context Menu

hojjatabdollahi:
The idea is that adding `.selectable()` to any `Text` should:
- make the text selectable using mouse
- user should be able to right click and choose select all/copy from a context menu
- the text should become focusable and the user should be able to navigate to it, select using shift+arrow keys and copy using ctrl+c

Also all `TextInput` and `TextEditor` widgets should automatically have the context menu (copy/cut/paste/select all) too. The user shouldn't need to add any special code to their `update()` for any of the widgets.

I have achieved the goals above with this implementation.

- [x] test wayland popup
- [x] test overlay (resizing the window crashed in x11, but it's a pre-existing issue)
- [x] merge requirements into Iced: https://github.com/pop-os/iced/pull/339
- [x] The selected_text_fill color is not being set correctly for Text
- [x] Implement the logic for the correct position for the context menu (right now it's always to the bottom right, causing it to clip in x11 and stick out needlessly on wayland)
- [x] The wayland popup context menu doesn't highlight the item on hover correctly, or consistently. I worry it's a deeper issue.

Still working on it. But I thought this is good enough in case there are any feedback.

The diff is kinda big, but two large files are `text` and `text_editor` which are just wrappers. 

___

### DIFF
--- a/examples/multi-window/src/window.rs
+++ b/examples/multi-window/src/window.rs
@@ -5,7 +5,7 @@ use cosmic::iced::core::{id, Alignment, Length, Point};
 use cosmic::iced::widget::{column, container, scrollable, text};
 use cosmic::iced::{self, event, window, Subscription};
 use cosmic::prelude::*;
-use cosmic::widget::{button, header_bar};
+use cosmic::widget::{button, header_bar, selectable_text};
 
 #[derive(Debug, Clone, PartialEq)]
 pub enum Message {
@@ -132,10 +132,14 @@ impl cosmic::Application for MultiWindow {
         let new_window_button = button::custom(text("New Window")).on_press(Message::NewWindow);
 
         let content = scrollable(
-            column![input, new_window_button]
-                .spacing(50)
-                .width(Length::Fill)
-                .align_x(Alignment::Center),
+            column![
+                input,
+                new_window_button,
+                selectable_text::title1("I'm selectable...")
+            ]
+            .spacing(50)
+            .width(Length::Fill)
+            .align_x(Alignment::Center),
         );
 
         let window_content = container(container(content).center_x(Length::Fixed(200.)))
@@ -146,7 +150,14 @@ impl cosmic::Application for MultiWindow {
         if id == self.core.main_window_id().unwrap() {
             window_content.into()
         } else {
-            column![header_bar().focused(focused), window_content].into()
+            container(column![
+                header_bar()
+                    .on_close(Message::CloseWindow(id))
+                    .focused(focused),
+                window_content
+            ])
+            .class(cosmic::theme::Container::WindowBackground)
+            .into()
         }
     }
 
--- a/src/app/cosmic.rs
+++ b/src/app/cosmic.rs
@@ -513,7 +513,7 @@ where
         &mut self,
         message: crate::Action<T::Message>,
     ) -> iced::Task<crate::Action<T::Message>> {
-        let message = match message {
+        let mut task = match message {
             crate::Action::App(message) => self.app.update(message),
             crate::Action::Cosmic(message) => self.cosmic_update(message),
             crate::Action::None => iced::Task::none(),
@@ -533,7 +533,7 @@ where
         #[cfg(all(target_env = "gnu", not(target_os = "windows")))]
         crate::malloc::trim(0);
 
-        message
+        task
     }
 
     #[cfg(not(feature = "multi-window"))]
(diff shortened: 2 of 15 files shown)


mmstick:
Merged the changes in iced and updated libcosmic.

hojjatabdollahi:
Still working on this. Not sure why I can't get the wayland popup context menu items to react to mouse hover consistently. I'll figure it out.

wash2:
I don't love the way that the context menu is created, but I understand that it makes this more of a drop-in change and support for the context menu less cumbersome in general. This creates the popup directly through sctk, but I wonder if it could use the surface `Action` instead, and maybe the `simple_popup` helper.

That aside, for `set_current_window_id`, we set the window id in the view method, but would it make sense to set this in response to a window focus event instead?

As for whether it is worth it to try to keep the best effort backward compatibility, I think that it generally is, but we could try to time this along with the changes for theme-v2, which also brings some slightly breaking changes.

hojjatabdollahi:
I'm working on another approach without needing to go to sctk directly. I'll push that soon.
It took me 2 days to figure out why the menu items aren't highlighting in wayland popups.
I'll have something soon.

hojjatabdollahi:
@wash2 I'm not sending SCTK actions directly anymore. Instead I'm using a subscription that sends a None Action wake message (instead of a SurfaceAction). This way the user doesn't need to handle SurfaceActions, and we can add context menu to all text inputs for free (see the diff here: https://github.com/pop-os/cosmic-files/pull/1815/changes). 

What do you think?


EDIT: I also updated the API, instead of `.selectable()` at the end of `Text` now you have to use `SelectableText`. This way the type of Text doesn't change and updating to a new libcosmic won't create a bunch of extra compile time errors. Instead for the widgets that you want to be selectable, instead of adding `.selectable()` at the end you change `Text` to `SelectableText`, ow `text::body()` to `selectable_text::body()`.

wash2:
> @wash2 I'm not sending SCTK actions directly anymore. Instead I'm using a subscription that sends a None Action wake message (instead of a SurfaceAction). This way the user doesn't need to handle SurfaceActions, and we can add context menu to all text inputs for free (see the diff here: https://github.com/pop-os/cosmic-files/pull/1815/changes).

I like this approach much more, and in the future, as long as we are taking this approach anyways, I'm wondering if we should refactor the handling of surface actions to function this way. 

> EDIT: I also updated the API, instead of `.selectable()` at the end of `Text` now you have to use `SelectableText`. This way the type of Text doesn't change and updating to a new libcosmic won't create a bunch of extra compile time errors. Instead for the widgets that you want to be selectable, instead of adding `.selectable()` at the end you change `Text` to `SelectableText`, ow `text::body()` to `selectable_text::body()`.

Thanks!

hojjatabdollahi:
> I'm wondering if we should refactor the handling of surface actions to function this way

I would say so. Cosmic Application trait should just give you these plumbings for free.

wash2 (dismissed):
This needs to be rebased now after the theme-v2 update, but it seems pretty good to me.

hojjatabdollahi:
- [ ] Fix issue with multi-window apps

trust2k:
I would like to request a similar task for opening and saving files (xdg-desktop-portal-cosmic), which is equally important. It often requires not only creating folders but also dragging and dropping files, copying and pasting them, renaming them, deleting them, or opening them in another program, as well as archiving them. Here is a layout of how it is supposed to look (taken from xdg-desktop-portal-lxqt). Thank you in advance!

Screenshot_2026-07-02_13-28-53

https://github.com/pop-os/xdg-desktop-portal-cosmic/issues/221

wash2 (dismissed):

hojjatabdollahi:
The issue I was seeing with multi window was opening the context menu in the child window will open it in the parent. 

I just tested `cargo run -p multi-window`, I see more issues:
- It's flickering
- Can't drag or right click on the header bar in the child window
- Context menu opens in the parent still



https://github.com/user-attachments/assets/636b1ffb-0f6e-48a9-b614-6f647c63f50c

hojjatabdollahi:
@wash2 The issue is probably the way the multi-window example is setup. Maybe the example needs to be updated.

wash2:
> @wash2 The issue is probably the way the multi-window example is setup. Maybe the example needs to be updated.

OK, I'll take another look at this today.

wash2 (approved):

PR #1307, opened 2026-06-09
Fix progress bar showing incorrectly with small width.

Aadil127:

https://github.com/user-attachments/assets/0efa09fc-0b41-4e28-acfa-3967eee8e443

Fix for this pop-os/cosmic-monitor#8

Need feedback on UI if this is intended design for progress on low value.

### DIFF
--- a/src/widget/progress_bar/linear.rs
+++ b/src/widget/progress_bar/linear.rs
@@ -197,33 +197,60 @@ where
         let border_color = custom_style.border_color.unwrap_or(custom_style.bar_color);
         let radius = custom_style.border_radius;
 
-        let mut draw_quad =
-            |x: f32, width: f32, color: iced::Color, border: iced::Border, marker_segment: bool| {
-                let height = if width * bounds.width
-                    < border.radius.top_left.max(border.radius.bottom_left)
-                    && !marker_segment
-                {
-                    (bounds.height - (2.0 * radius).min(bounds.height) + width * bounds.width * 2.0)
-                        .min(bounds.height)
+        let mut draw_quad = |x: f32,
+                             width: f32,
+                             color: iced::Color,
+                             mut border: iced::Border,
+                             is_track: bool,
+                             total_progress_width: f32| {
+            let mut height = bounds.height;
+            if !is_track {
+                // For progress that is at the end of completion
+                if total_progress_width > bounds.width - radius {
+                    let border_radius =
+                        radius.min(bounds.height / 2.0) - (bounds.width - total_progress_width);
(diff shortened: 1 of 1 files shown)


Aadil127:
https://github.com/user-attachments/assets/ee75d5a9-5ecf-435f-9987-02f4fd59ef8d


Fix in second commit - progress bar with different height (girth) behaves similarly on low progress phase.
I would like feedback on it.

maria-komarova:
Looking at the two recordings, the right edge of the active track should be flat and not rounded. If it would be possible to make it look like the active track is filling the entire space available inside the progress bar at low level that would be better. Right now it looks like it has the rounded edge.

Aadil127:
**After Fix**

https://github.com/user-attachments/assets/9ae43284-60b6-4b55-b017-b5c53fdd6fef


https://github.com/user-attachments/assets/a78887c8-fa81-49fe-96b9-129667d3814a

jackpot51 (approved):

git-f0x:
The rounded bar with markers on the top in the recording above seems to have the bar going outside track bounds at the end, unlike the one below it.
All bars also still go a bit outside bounds at the start (especially for fully rounded tracks; e.g. cpu5), and can have vertical alignment issues, probably because of pixel alignment (bottom right bar):
Screenshot_20260610_173227

A proper fix would likely need some way of masking things going outside a "parent" widget in iced, so that the bar could be fully rectangular, and the track masks it to match it.

Aadil127:
We can do gradient for one quad the right side of gradient can have track color and left can have progress bar color.

```
let background = Background::Gradient(iced::Gradient::Linear(
                iced_core::gradient::Linear::new(90.0_f32.to_radians()).add_stops([
                    iced_core::gradient::ColorStop {
                        color: custom_style.bar_color,
                        offset: 0.0,
                    },
                    iced_core::gradient::ColorStop {
                        color: custom_style.bar_color,
                        offset: mid_point,
                    },
                    iced_core::gradient::ColorStop {
                        color: custom_style.track_color,
                        offset: mid_point + f32::EPSILON,
                    },
                    iced_core::gradient::ColorStop {
                        color: custom_style.track_color,
                        offset: 1.0,
                    },
                ]),
            ));
```
Something like this so instead of smooth gradient it will abruptly switch color so only mid_point (offset for center colors) needs to be calculated.

Aadil127:
https://github.com/user-attachments/assets/f70e8e14-b7fd-43d3-991c-d79288d08520

Indeterminate progress bar and progress bar with marker are currently left.

PR #1311, opened 2026-06-11
Fix : using gradient to render progrees bar to prevent overflow.

Aadil127:
Fixed progress bar overflow.
Now even for low progress phase the active progress bar does not go outside of that track.
Are there any changes needed?

https://github.com/user-attachments/assets/602721ac-45ee-40ab-a71a-e48be41a0d70

### DIFF
--- a/src/widget/progress_bar/linear.rs
+++ b/src/widget/progress_bar/linear.rs
@@ -201,25 +201,24 @@ where
 
         let mut draw_quad =
             |start: f32, segment_width: f32, progress_width: f32, border: iced::Border| {
-                let background = if progress_width == 0.0 {
-                    iced::Background::Color(custom_style.track_color)
-                } else if (segment_width - progress_width).abs() < f32::EPSILON {
-                    iced::Background::Color(custom_style.bar_color)
-                } else {
-                    let mid_point = progress_width / segment_width;
-                    iced::Background::Gradient(iced::Gradient::Linear(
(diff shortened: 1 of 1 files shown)


git-f0x:
Can you restore `radius_inner`? The inner segments shouldn't be fully rectangular.
Some things also feel a bit weird, e.g:
```
for _ in 0..6 {
    left_width_offset = left_width_offset.next_down();
}
```
but I never worked with gradients, so not sure if that might be needed.

Aadil127:
https://github.com/user-attachments/assets/3ff68bf7-e64c-4a27-852f-33724d2cb0a3

Currently for marker segments if that segment is empty or full it directly uses that color for background instead of gradient but for segment where it is using gradient it's edges becomes blurry.
So marker does not stay sharp.

**Current one**
current

**Approach using gradient**
If i use gradient for all of the segments.
gradient
gradient one is slightly blurry.

Aadil127:
color_background<br>
gradient_backgroud

When segment switches from color background to gradient background it creates 1 pixel wide blur line (see left and right markers) 
Using floor() on horizontal line seems to fix blur for that but it can not be done for vertical marker lines because some marker can appear wider than intended.

Aadil127:
https://github.com/user-attachments/assets/5c6aa524-4d21-4f74-bdb7-6e0d22f529be

**This is currently how it looks.**

Aadil127:
Closing this because found new approach to solve it.
Here's new PR for that #1314

PR #1339, opened 2026-07-08
fix: no text should use the steps array for transparent color

wash2:

### DIFF
--- a/cosmic-theme/src/model/theme.rs
+++ b/cosmic-theme/src/model/theme.rs
@@ -11,7 +11,6 @@ use palette::color_difference::Wcag21RelativeContrast;
 use palette::rgb::Rgb;
 use palette::{IntoColor, Oklcha, Srgb, Srgba, WithAlpha};
 use serde::{Deserialize, Serialize};
-use std::default;
 use std::num::NonZeroUsize;
 
 /// ID for the current dark `ThemeBuilder` config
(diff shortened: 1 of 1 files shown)


hojjatabdollahi:
This fixes the text color issue. Are the text in the context drawer supposed to be dim?
image

wash2:
Thanks again for catching that

hojjatabdollahi:
Thanks. Looks good.

I noticed 2 things:
1- the context drawer sometimes is blurred sometime isn't depending on the app:
image

2- when we're max thickness, min opacity the container background becomes transparent, making the UI a bit weird, but I assume that's by design?
image

compared to a config with more mid opacity for example (still max thickness):
image

wash2:
> Thanks. Looks good.
> 
> I noticed 2 things: 1- the context drawer sometimes is blurred sometime isn't depending on the app: image
> 
> 2- when we're max thickness, min opacity the container background becomes transparent, making the UI a bit weird, but I assume that's by design? image
> 
> compared to a config with more mid opacity for example (still max thickness): image

1. This is a result of not having a good way to implement blur in the context drawer when it is overlayed on top of the app, rather than adjacent. I think that neither popups nor subsurfaces are appropriate.
2. Ya, this is kinda by design when dragging the sliders to the extremes. It defaults to a reasonable value, which is probably fine.

jackpot51 (approved):

PR #1342, opened 2026-07-10
feat: add capture_guard widget

RipleyTom:
This adds a capture_guard widget to libcosmic whose goal is to avoid passing on events that have been captured to children so that they do not lose their focus. This was made specifically for libcosmic applications losing their focus when moved by clicking the title bar or resized.

The code is very similar to id_container and I've thought about maybe merging them but that'd add some conditions at execution and a bit of spaghetti so I'm not sure.

### DIFF
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -858,7 +858,7 @@ impl<App: Application> ApplicationExt for App {
                 None
             })
             // The content element contains every element beneath the header.
-            .push(content)
+            .push(crate::widget::capture_guard(content))
             .apply(container)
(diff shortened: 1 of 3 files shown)


RipleyTom:
Hmm git-f0x pointed out to me that libcosmic application can add widgets to the title bar, ie before main content through header_start, etc(which I missed because I was focused on cosmic-term and cosmic-edit which don't have those).
So the current approach isn't good enough.

RipleyTom:
Yeah nvm doing anything else would require adding a variable to Shell in iced to keep track of which widget handled the event and ~eh, too messy.

PR #1362, opened 2026-07-15
feat: Add option for frozen glass in maximized windows

### LINKED ISSUE #1361: Maximized apps don't get frozen glass effect
I know that it's by design like this that when you maximize window the frozen glass effect is gone, but I really miss it when I'm maximizing apps. There are some workaround like setting borders to zero, disabling active window tint, but I find them useful when I'm having multiple apps open in one workspace. I think it would be best if it was configurable e.g. on theme level if you want to have frozen glass applied when the windows is maximized.

vacenty:

Resolves #1361. I'm missing frozen glass effect when apps are maximized, so I've added option to enable it as part of theme - I think it was the easiest to implement and probably makes the most sense. I guess if it's accepted I'd then try to update cosmic settings, so that it's configurable from GUI. I think this PR should be backwards compatible with current settings.

### DIFF
--- a/cosmic-theme/src/model/theme.rs
+++ b/cosmic-theme/src/model/theme.rs
@@ -121,6 +121,9 @@ pub struct Theme {
     /// frosted applet popups
     pub frosted_applets: bool,
     #[serde(default)]
+    /// frosted apps when maximized
+    pub frosted_maximized_apps: bool,
+    #[serde(default)]
     /// alpha map
     pub alpha_map: AlphaMap,
     /// shade color for dialogs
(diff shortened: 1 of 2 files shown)


Levo75:
That would be nice.

wash2 (approved):
This seems fine to me, though UX would have to approve any additional settings toggles. If not cosmic-settings, the toggle would probably be welcome in https://github.com/cosmic-utils/tweaks

vacenty:
Just in case it's accepted in cosmic settings I've added PR for it https://github.com/pop-os/cosmic-settings/pull/2130

PR #1370, opened 2026-07-23
fix(slider): add handle bounds to avoid handle shifting and apply correct corner radius.

Aadil127:
- [x] Depends on pop-os/iced#357
<hr>

### DIFF
--- a/src/theme/style/iced.rs
+++ b/src/theme/style/iced.rs
@@ -775,12 +775,12 @@ impl slider::Catalog for Theme {
 
                     handle: slider::Handle {
                         shape: slider::HandleShape::Rectangle {
-                            height: 20,
-                            width: 20,
+                            height: 26,
+                            width: 26,
                             border_radius: cosmic.corner_radii.radius_m.into(),
                         },
                         border_color: Color::TRANSPARENT,
-                        border_width: 0.0,
+                        border_width: 3.0,
                         background: Background::Color(cosmic.accent.base.into()),
                     },
 
@@ -795,12 +795,6 @@ impl slider::Catalog for Theme {
             slider::Status::Active => appearance,
             slider::Status::Hovered => match class {
                 Slider::Standard => {
-                    appearance.handle.shape = slider::HandleShape::Rectangle {
-                        height: 26,
-                        width: 26,
-                        border_radius: cosmic.corner_radii.radius_m.into(),
-                    };
-                    appearance.handle.border_width = 3.0;
                     appearance.handle.border_color =
                         self.cosmic().palette.neutral_10.with_alpha(0.1).into();
                     appearance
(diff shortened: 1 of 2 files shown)


src/theme/style/iced.rs
```diff
@@ -809,20 +803,9 @@ impl slider::Catalog for Theme {
...
-                            height: 26,
-                            width: 26,
-                            border_radius: cosmic.corner_radii.radius_m.into(),
-                        };
-                        appearance.handle.border_width = 3.0;
-                        appearance.handle.border_color =
-                            self.cosmic().palette.neutral_10.with_alpha(0.1).into();
-                        appearance
-                    };
-                    style.handle.border_color =
```
mmstick, line 823: This is altering the style of the handle

Aadil127: on active state handle's inner width is 20 with transparent border width of 3.
on hovered it changes the border color with 0.1 alpha
and on dragging it changes border color with 0.2 alpha of palette.neutral_10 color.

https://github.com/pop-os/iced/pull/357#issue-4711300561 this is the video after changes.

Is it intentional to have slider handle round even on slightly rounded style chosen in appearence?

previously handle width was 20 with 0 border width
and on hover and drag it changes the width to 26 with border width of 3.
Now handle have width of 26 with border width of 3 on all state and only changes the border color that was previously defined in theme.

mmstick: Yes, any changes to the style will require a round of UX reviews.

Aadil127:
Here's an illustration of changes made for UX team.

In newer change instead of slider rail spanning entire width of widget it starts at the center of handle bounds when it is on 0% progress and same for other side as well.
This avoids slider rail being visible on hover due to opaque border.

Also changed size of handle bounds to be consistent on all states(active, hover, drag) ,
height and width of 26 (20 for handle and 3 for border on each side) so border radius can be applied correctly.
Here's a video of that https://github.com/pop-os/iced/pull/357#issue-4711300561
start
mid

mmstick:
There shouldn't be any changes to the widget design. I guess I'll revert the changes in iced until this is fixed.

mmstick:
The slider looks fine without the style changes here so we can update dependencies without it.

Aadil127:
https://github.com/user-attachments/assets/5576644c-9203-4a2a-bd31-b61a29512d67

without updating style corner radius is inconsistant.
Currently in style of slider handle's width is 20 in active state and on drag and hover it's width is 26 so corner radius becomes inconstant without updating style.
(on slightly rounded style on active state handle looks more circular) 


**updated style**

https://github.com/user-attachments/assets/36ee958d-85cf-4d87-9153-0b0093ea9590

mmstick:
I see. I'll merge since the style looks the same as before.

mmstick (approved):

PR #1381, opened 2026-08-04
fix(menu_bar): fall back to overlay menus on non-COSMIC Wayland compositors

gaikwadyash905:
The menu bar assumed that any Wayland session supports COSMIC's native popup protocols (via cctk). On other compositors like GNOME/mutter, the popup creation silently fails while the overlay fallback is also disabled, resulting in menus that register clicks but never display.

Add a `use_wayland_popups()` helper that checks both the windowing system and `XDG_CURRENT_DESKTOP` to determine if the COSMIC compositor is running. When it is not, the menu bar falls back to the standard overlay-based rendering path that works on all compositors.

Ref pop-os/cosmic-edit#593


mmstick:
It should not matter what desktop environment you are using. There are no COSMIC-specific protocols being used here.

PR #1382, opened 2026-08-04
fix: request redraw 

hojjatabdollahi:
Previously iced was requesting redraw on mouse move, so some widgets looked correct without asking for a redraw explicitly. But with https://github.com/pop-os/iced/pull/377, the widgets need to ask for it explicitly. This is more correct and more efficient.

For example Button wasn't telling anybody that it needs a redraw when it would change its hover state, but since we were redrawing as the cursor moves, we would see the hover state change. Now, it asks explicitly. So, instead of getting a redraw per monitor refresh rate as the cursor moves, you only get 2, one when the cursor enters the button and the hover state changes, and one when it leaves the button.

____

### DIFF
--- a/src/widget/button/widget.rs
+++ b/src/widget/button/widget.rs
@@ -868,8 +868,21 @@ pub fn update<'a, Message: Clone>(
                 }
             }
         }
+        Event::Mouse(mouse::Event::CursorMoved { .. })
+        | Event::Touch(touch::Event::FingerMoved { .. }) => {
+            let state = state();
+            let is_hovered = cursor.is_over(layout.bounds());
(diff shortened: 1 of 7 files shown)


hojjatabdollahi:
8ad9a08 fixes libcosmic apps with responsive menubar sending 5k messages per seconds (my CPU can't handle more) when the window is smaller than the content. Iced requests redraw when a message arrives. So a small window keeps redrawing without anything changing.

https://github.com/user-attachments/assets/43848118-ea99-4f60-87ad-c14b4ce04f86

wash2 (approved):

PR #1390, opened 2026-08-13
Addition of choice of window controls to the left or to the right

JackNWhite:
This PR adds the apparatus for declaring in a COSMIC theme whether standard window controls should appear on the left or the right of the header bar:
- A boolean field in the `cosmic_theme::Theme` and `cosmic::ThemeBuilder` structs
- Initialisation code in the implementations of same
- A getter for the active theme in `theme`
- Determination of whether to add the controls left or right in the HeaderBar widget's `view()` function

If the controls are configured left, the order is `[close][min][max]` and if right then `[min][max][close]`.

## Checklist
- There is no AI-generated code
- I believe I understand what I have done
- My commit messages are accurate
- I have tested my contribution using the template application and I have run `just check` on my branch
- I have read and certify my contribution under the [Developer Certificate of Origin](https://developercertificate.org/)

### DIFF
--- a/cosmic-theme/src/model/theme.rs
+++ b/cosmic-theme/src/model/theme.rs
@@ -99,6 +99,8 @@ pub struct Theme {
     pub is_dark: bool,
     /// is high contrast
     pub is_high_contrast: bool,
+    /// are standard window controls on the left
+    pub window_controls_left: bool,
     /// cosmic-comp window gaps size (outer, inner)
     pub gaps: (u32, u32),
     /// cosmic-comp active hint window outline width
@@ -891,6 +893,8 @@ pub struct ThemeBuilder {
     #[serde(default)]
     /// enabled blurred transparency
     pub frosted: BlurStrength,
+    /// standard window controls are set left
+    pub window_controls_left: bool,
     /// cosmic-comp window gaps size (outer, inner)
     pub gaps: (u32, u32),
     /// cosmic-comp active hint window outline width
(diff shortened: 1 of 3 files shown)


mmstick:
That should be a [CosmicTK](https://github.com/pop-os/libcosmic/blob/master/src/config/mod.rs) config instead of a theme parameter.

JackNWhite:
> That should be a [CosmicTK](https://github.com/pop-os/libcosmic/blob/master/src/config/mod.rs) config instead of a theme parameter.

OK, thank you, @mmstick! I will look at that over the weekend. Could you tell me though, why this should be a CosmicTk config but dark mode or frosting is in the theme?

mmstick:
The frosted effect is managed on a per-theme basis and there are different themes for dark and light modes. Different themes may use different values for frosted glass and the dark theme parameter affects how the theme is programmatically generated and applied.

CosmicTk is used for configuring the toolkit globally, regardless of the chosen theme, such as the system font, system icon theme, whether global themes are enabled, and window controls. This already has configuration settings for toggling the minimize and maximize buttons, so a setting to hypothetically alter the position of the window controls should also be here.

That said, you'll need UX design approval to add this since it affects the design of all applications. I'm not sure if this is something we're ready to do right now.

sousaakira:
I think this would be a great addition to COSMIC. 👍

I agree with the suggestion of making this a CosmicTK setting, since the position of window controls is a global desktop preference rather than something that necessarily belongs to the theme.

It would be really useful to have this option available in the system settings, allowing users to choose whether the window controls appear on the left or right. This would also make COSMIC more flexible for users coming from different desktop environments who are already accustomed to a particular button layout.

Thank you for working on this! I hope the UX team considers the proposal. It would be a great customization option for COSMIC.

JackNWhite:
Since I will be reverting my changes in this PR anyway, I am closing it and will reference it in another.

PR #1396, opened 2026-08-17
fix: request redraw on dropdown open/close in the in-window overlay path

### LINKED ISSUE #1395: widget::dropdown popup shows nothing until the mouse moves out of the dropdown area
## Summary
On Wayland, opening a `cosmic::widget::dropdown` (via click) creates a real `xdg_popup` surface for the option list, but nothing forces that surface to paint its first frame. The popup sits blank/invisible until the pointer generates a `CursorMoved` event over it - at which point it suddenly appears. Moving the mouse immediately after clicking "fixes" it, which makes it easy to miss in casual testing but very noticeable for
anyone who clicks and then reads before moving the mouse.

## Environment
- `libcosmic` pinned at `1f8d8786940cc6db59d412153c5a03f69a20375c` (currently == tip of `master`)
- COSMIC session (`cosmic-comp`), Wayland, `wayland_platform` cfg active
- Reproduced with a plain `cosmic::widget::dropdown(labels, selected, on_selected)`, no custom styling or wrapping — this isn't specific to any one consuming app

## Steps to reproduce
1. Build any `libcosmic` app with a `widget::dropdown` on screen.
2. Click the dropdown to open it, and **do not move the mouse afterward**.
3. Observe: the popup surface is created (compositor treats it as mapped — clicking elsewhere closes it, keyboard/other input behaves as if it's open) but nothing is drawn — the menu options are not visible.
4. Move the mouse by even one pixel over the popup area.
5. Observe: the menu now renders correctly.

## Expected behavior
The popup should render its option list on the same frame it's created/mapped, with no dependency on a subsequent pointer-motion event.

sidechannelio:
`Dropdown::new` (the plain constructor used by any non-applet consumer) leaves `window_id`/`on_surface_action` unset, which routes the widget's `overlay()` impl to the in-window overlay path rather than the surface-action-based popup path. On that path, `update()`'s `open` closure and its close counterparts only ever requested a redraw inside the `#[cfg(wayland_platform)] if let Some(...)` block gated on those fields being set — so for the common case, opening or closing a dropdown flipped `is_open` with no redraw requested at all. The stale frame stuck around until some unrelated later event (e.g. cursor motion landing on the now-open/closed menu's bounds) forced a repaint, making the menu appear to not open until the mouse moved, and not close until the same thing happened again on the way out.

Add an unconditional `shell.request_redraw()` at each of the three `is_open` transitions that don't already publish a Message of their own (an actual option selection already triggers a normal Message-driven redraw through the app's own update cycle, so that path needed no change):

- `open`: right after `is_open` is set true, ahead of the wayland-only surface-action block.
- the `close_operation` branch of the open/close operation state machine (`Id`-based programmatic close).
- the click-outside-closes-it branch of the mouse/touch press handler.

Fixes #1395

### DIFF
--- a/src/widget/dropdown/widget.rs
+++ b/src/widget/dropdown/widget.rs
@@ -661,6 +661,7 @@ pub fn update<
         state.close_operation = false;
         state.is_open.store(false, Ordering::SeqCst);
         if is_open {
+            shell.request_redraw();
             #[cfg(wayland_platform)]
             if let Some(ref on_close) = on_surface_action {
                 shell.publish(on_close(surface::action::destroy_popup(state.popup_id)));
@@ -684,6 +685,7 @@ pub fn update<
                 // Event wasn't processed by overlay, so cursor was clicked either outside it's
                 // bounds or on the drop-down, either way we close the overlay.
                 state.is_open.store(false, Ordering::Relaxed);
+                shell.request_redraw();
                 #[cfg(wayland_platform)]
                 if let Some(on_close) = on_surface_action {
                     shell.publish(on_close(surface::action::destroy_popup(state.popup_id)));


hojjatabdollahi:
@iampaulmata I cherrypicked your commit and created a new PR with some more fixes for dropdown widget. I hope you don't mind.

PR #1406, opened 2026-08-26
feat: support window hint alpha

abachmann:
Change `window_hint` type from `Srgb` to `Srgba` to allow users to specify an alpha value.

---

### DIFF
--- a/cosmic-theme/src/model/theme.rs
+++ b/cosmic-theme/src/model/theme.rs
@@ -104,7 +104,7 @@ pub struct Theme {
     /// cosmic-comp active hint window outline width
     pub active_hint: u32,
     /// cosmic-comp custom window hint color
-    pub window_hint: Option<Srgb>,
+    pub window_hint: Option<Srgba>,
     #[serde(default)]
     /// enables blurred transparency
     pub frosted: BlurStrength,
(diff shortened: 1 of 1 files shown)


cosmic-theme/src/model/theme.rs
```diff
@@ -104,7 +104,9 @@ pub struct Theme {
     /// cosmic-comp active hint window outline width
     pub active_hint: u32,
     /// cosmic-comp custom window hint color
-    pub window_hint: Option<Srgb>,
+    #[serde(with = "color_serde_option")]
+    #[cosmic_config_entry(with = ColorReprOption)]
```
mmstick: This would break all existing v2 themes since it changes the format from `(red: f32, blue: f32, green: f32)` to a hex string. The existing parser can happily accept an alpha parameter to the struct tuple above so this format change isn't needed.

mmstick (changes requested):

mmstick:
Is there a reference issue that is approved by UX?

cosmic-theme/src/model/theme.rs
abachmann: done

abachmann:
> Is there a reference issue that is approved by UX?

Not that I'm aware of. I just played around with the window_hint setting and thought it would be nice to make the alpha value also configurable.

PR #1412, opened 2026-09-02
perf(cosmic-config): batch theme write transactions

ByteAtATime:
## Problem

`ConfigTransaction::commit` performs two fsync operations per file, which is really slow; a single theme change (~60ish keys) takes 183 fsyncs and 630ms.

This was noticed while working on [cosmic-theme-import](https://github.com/ByteAtATime/cosmic-theme-import), which takes ~1 second without this patch and about half a second with it.

## Results

Theme change:

| | master | this patch |
|---|---|---|
| fsync count | 183 | 62 (61 files + 1 dir) |
| commit latency | ~610 ms | ~310 ms |

### DIFF
--- a/cosmic-config/src/lib.rs
+++ b/cosmic-config/src/lib.rs
@@ -461,18 +461,73 @@ pub struct ConfigTransaction<'a> {
 
 impl ConfigTransaction<'_> {
     /// Apply all pending changes from ConfigTransaction
-    //TODO: apply all changes at once
     pub fn commit(self) -> Result<(), Error> {
-        let mut updates = self.updates.lock().unwrap();
-        for (key_path, data) in updates.drain(..) {
-            atomicwrites::AtomicFile::new(
-                key_path,
-                atomicwrites::OverwriteBehavior::AllowOverwrite,
-            )
-            .write(|file| file.write_all(data.as_bytes()))?;
-        }
-        Ok(())
+        let updates = self.updates.lock().unwrap().drain(..).collect::<Vec<_>>();
+        write_updates(updates).map_err(Error::Io)
     }
+}
+
+/// Atomically write all updates, using one fsync per staged file and one fsync
+/// per distinct parent directory instead of one transaction per key.
+fn write_updates(updates: Vec<(PathBuf, String)>) -> std::io::Result<()> {
+    if updates.is_empty() {
+        return Ok(());
+    }
+
+    let pid = std::process::id();
+    let mut staged: Vec<(PathBuf, PathBuf)> = Vec::with_capacity(updates.len());
+
+    let staged_result = updates
+        .iter()
+        .enumerate()
+        .try_for_each(|(index, (key_path, data))| {
+            let dir = key_path.parent().ok_or_else(|| {
+                std::io::Error::new(
+                    std::io::ErrorKind::InvalidInput,
+                    "key path has no parent directory",
+                )
+            })?;
+            let temp_path = dir.join(format!(".cosmic-config-{pid}-{index}.tmp"));
+            let mut file = fs::OpenOptions::new()
+                .write(true)
+                .create_new(true)
+                .open(&temp_path)?;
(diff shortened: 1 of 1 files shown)


mmstick:
Note that you can export and import themes natively with `cosmic-settings appearance import {{theme.ron}}` and `cosmic-settings appearance export {{theme.ron}}`.

ByteAtATime:
I am aware, I just wanted to make a small binary as I am not on COSMIC and would prefer to not install the full cosmic-settings package.

cosmic-config/src/lib.rs
```diff
@@ -461,18 +461,73 @@ pub struct ConfigTransaction<'a> {
...
-    //TODO: apply all changes at once
     pub fn commit(self) -> Result<(), Error> {
-        let mut updates = self.updates.lock().unwrap();
-        for (key_path, data) in updates.drain(..) {
-            atomicwrites::AtomicFile::new(
-                key_path,
-                atomicwrites::OverwriteBehavior::AllowOverwrite,
-            )
-            .write(|file| file.write_all(data.as_bytes()))?;
+        let updates = self.updates.lock().unwrap().drain(..).collect::<Vec<_>>();
```
mmstick: The previous code holds the lock while writing while this drops the lock before writing.

cosmic-config/src/lib.rs
```diff
@@ -461,18 +461,73 @@ pub struct ConfigTransaction<'a> {
...
+                    "key path has no parent directory",
+                )
+            })?;
+            let temp_path = dir.join(format!(".cosmic-config-{pid}-{index}.tmp"));
+            let mut file = fs::OpenOptions::new()
+                .write(true)
+                .create_new(true)
+                .open(&temp_path)?;
+            file.write_all(data.as_bytes())?;
+            file.sync_all()?;
```
mmstick: If multiple files are being written in one update then it would be better to perform the sync after the writes have been submitted for the temporary files. File descriptors could be kept to avoid needing to reopen them.

ByteAtATime:
oops - screwed up a bit with the history but I think it should be fine

cosmic-config/src/lib.rs
ByteAtATime: Fixed in [`1664605`](https://github.com/pop-os/libcosmic/pull/1412/changes/1664605c32dd5ccd5db87fbfc5ecd1a7b9b0e03d)

cosmic-config/src/lib.rs
ByteAtATime: Fixed in [`ed0b9b6`](https://github.com/pop-os/libcosmic/pull/1412/commits/ed0b9b6a23779f21fbbe467d283f7e95934f1a6a). I don't think it makes a difference in this specific change/use case but it's probably good practice

cosmic-config/src/lib.rs
```diff
@@ -461,18 +461,65 @@ pub struct ConfigTransaction<'a> {
...
+            staged.push((file, temp_path, key_path));
+            Ok(())
+        });
+
+    let result = staged_result.and_then(|()| {
+        staged.iter().try_for_each(|(file, _, _)| file.sync_all())?;
+
+        staged
+            .iter()
+            .try_for_each(|(_, temp_path, key_path)| fs::rename(temp_path, key_path))?;
```
mmstick, line 500: It would be better to perform the renames the same as the previous solution: https://docs.rs/atomicwrites/latest/src/atomicwrites/lib.rs.html#179-206

mmstick: And on Windows: https://docs.rs/atomicwrites/latest/src/atomicwrites/lib.rs.html#314-323

ByteAtATime: That would double the fsyncs, which reverts the performance gains (300ish ms back to 600ish). The root problem is the fact that it's spread across a bunch of files, and I think the performance is just unavoidable - I'll close the PR.

PR #1416, opened 2026-09-02
context_menu fixes

hojjatabdollahi:
- `context_menu` takes an `Option` for its menu, but `diff` unwrapped it, panicking whenever `None` was passed. Skip the menu tree diff when there is no menu (how the other methods already handle it).
- Fix menu position in X11 mode. Add a 1x1 rectangle to anchor overlay context_menu to.
- The menu width (for both wayland and x11) is hardcoded to 240. Make it configurable.
- Add ability to reserve the icon space for a menu entry. (Needed for menu items in cosmic-files's context menu, the sort items). This is backward compatible, so nothing should break, but you can add a `reserver_icon()` to a menu item, to indent it and align it with its sibling that do have an icon.
- report open/close of context_menu, this is then used to highlight the item being right-clicked.
- for segmented button, report the correct live menu, so the item is highlighted correctly.
- change `surface::Action` to be typed on the Message. So, nested widgets inside components can have context menu with custom View.
- previously one id was being reused for each menu popup, but since now we receive a "Done" when pop up closes, to avoid racy bugs, it uses unique IDs for all pop ups. (Not sure what the consequences of this is yet).

Depends on:
- [x] https://github.com/pop-os/iced/pull/389

**NOTE: This is a breaking change, and every libcosmic app that needs Surface Actions needs to be updated.**

I have a few PRs ready for `cosmic-edit`, `cosmic-term`, `cosmic-files`, and `cosmic-viewer`. Mostly because they used to use a combination of wayland and overlay popups for context menus, or had implemented things locally.

___

### DIFF
--- a/examples/applet/src/window.rs
+++ b/examples/applet/src/window.rs
@@ -35,7 +35,7 @@ pub enum Message {
     PopupClosed(Id),
     ToggleExampleRow(bool),
     Selected(usize),
-    Surface(cosmic::surface::Action),
+    Surface(cosmic::surface::Action<Message>),
     Toggle(bool),
 }
 
@@ -76,9 +76,7 @@ impl cosmic::Application for Window {
                 self.example_row = toggled;
             }
             Message::Surface(a) => {
-                return cosmic::task::message(cosmic::Action::Cosmic(
-                    cosmic::app::Action::Surface(a),
-                ));
+                return cosmic::task::message(cosmic::Action::Surface(a));
             }
             Message::Selected(i) => {
                 self.selected = Some(i);
--- a/examples/application/src/main.rs
+++ b/examples/application/src/main.rs
@@ -98,7 +98,7 @@ pub enum Message {
     Input2(String),
     Ignore,
     ToggleHide,
-    Surface(cosmic::surface::Action),
+    Surface(cosmic::surface::Action<Message>),
     Hi,
     Hi2,
     Hi3,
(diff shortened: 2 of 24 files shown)


hojjatabdollahi:
@wash2 `surface::Action` was storing the view as `Box<dyn any>` and downcast it to the application message at render time.

But, a widget nested in a component with its own message type produced view of the wrong type, the downcast would silently fail and the view would show the view_window.

For example, in cosmic-files, the messages inside the tab, are typed to that, so context menu inside the tab would fail. That's why we were re-implementing everything inside cosmic-files, instead of using libcosmic's context menu.

So in [b040a91](https://github.com/pop-os/libcosmic/pull/1416/commits/b040a91b558bcd5fbce0bb4089ec5ae59ac39d55) I'm making `surface::Action` generic over the message type. The view is now stored as `View<M>` instead of `BoxedSetting`, then `Action::map` re-types an action the way `Element::map` re-types an element.

This is a breaking change for apps that have a wayland context menu. For example `cosmic-viewer`: [49bb982](https://github.com/pop-os/cosmic-viewer/pull/49/commits/49bb98235905d5c7918469a66f6223221967ac6e).

What do you think? Is there something I didn't think of?

wash2 (dismissed):

hojjatabdollahi:
I think tool tips in Cosmic files are not working anymore. Let me see.

mmstick (approved):

PR #1427, opened 2026-09-10
Fix text selection color

hojjatabdollahi:
Looks like https://github.com/pop-os/libcosmic/pull/1409 is dividing the text into 3 chunks of not-selected, selected, and not-selected and redrawing the graphemes, that doesn't work for mixed bidi-text. Also another bug is causing the text to be drawn with the wrong y.

This fixes the issue, by redrawing the selected text over the original text and clipping it to the selection quads.

Before:
Screenshot_2026-09-10_11-05-14

Screenshot_2026-09-10_11-58-14

After:

Screenshot_2026-09-10_11-59-15

Alternative approach is to set the color of the text span and let cosmic-text handle it, but that means everything should become rich-text and that's more expensive than redrawing the text over the original one.

____

### DIFF
--- a/src/widget/text_input/input.rs
+++ b/src/widget/text_input/input.rs
@@ -2920,27 +2920,39 @@ pub fn draw<'a, Message>(
             text_color
         };
 
-        renderer.fill_text(
-            Text {
-                content: if text.is_empty() {
-                    placeholder.to_string()
-                } else {
-                    text.clone()
-                },
-                font,
-                bounds: bounds.size(),
-                size: iced::Pixels(size),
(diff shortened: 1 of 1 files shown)


mmstick (approved):

xelavopelk:
Thank you for the error cases. I considered the option of redrawing the text. I decided to take a more complex route by splitting the text into three parts. The reason for this is that when redrawing, the pixels of the old text are sometimes visible. However, I was wrong; my solution was more fragile.

Thanks  @mmstick for quick rollback my buggy PR.

PR #1436, opened 2026-09-15
[Progress bar] Custom styling

git-f0x:
This adds the ability to set custom values for the progress_bar style fields. Also migrates it to use `Catalog` for style, instead of the older syntax.

I noticed that the printer setup work uses custom containers to color the progress bar a different color, which end up not matching other progress bars. With this, a progress bar can be colored magenta with:
`.class(Class::custom().bar_color(Color::from_rgb8(255, 0, 255)))`

CC @Abd002


git-f0x:
I removed the Class enum (it's just an override struct now), since it was just falling back to (computed) default values anyways. And the API ends up almost the same (`.class(Class::default().bar_color(Color::from_rgb8(255, 0, 255)))`).

I've also added helpers to both widgets so using .class() isn't needed. So setting magenta can be just `.bar_color(Color::from_rgb8(255, 0, 255)))` on the widget.

mmstick (approved):

